PR #2, opened 2021-07-19
Update all score processors and tests to use new v2 storage format

peppy:
Don't really need review on this, just PRing in its "test passing" state. Will be deploying this as part of an ongoing process later today or tomorrow.


smoogipoo:
Just pointing out that force pushes happened and i'm not sure on the state of this branch as a result.

smoogipoo:
Structure looks good.

peppy:
> Just pointing out that force pushes happened and i'm not sure on the state of this branch as a result.

Was just rebases on stuff i moved to `master` for immediate deployment. v2 stuff is a bit pending until osu-web stuff is also ready to go.

smoogipoo (approved):

PR #4, opened 2021-09-21
Add flashlight difficulty attribute

smoogipoo:
As a result of https://github.com/ppy/osu/pull/14217. Requires nuget package update.


smoogipoo:
Only merge this one after a new diffcalc run has been done, otherwise the pp processor will spew errors.

peppy (approved):

PR #6, opened 2021-10-05
Add user total performance points processor

smoogipoo:
- [x] Depends on https://github.com/ppy/osu-queue-score-statistics/pull/8
- Shouldn't be merged/deployed until the final switch to solo_scores (from the `osu_scores_high` tables), and until osu-performance is decommissioned.
- Should not consider lazer-made scores (needs another column in the table).

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserTotalPerformanceProcessor.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserTotalPerformanceProcessor.cs
@@ -1,0 +1,77 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using System.Linq;
+using Dapper;
+using MySqlConnector;
+using osu.Server.Queues.ScoreStatisticsProcessor.Models;
+
+namespace osu.Server.Queues.ScoreStatisticsProcessor.Processors
+{
+    public class UserTotalPerformanceProcessor : IProcessor
+    {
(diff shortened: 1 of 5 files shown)


osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserTotalPerformanceProcessor.cs
```diff
@@ -0,0 +1,81 @@
...
+            // Build the diminishing sum
+            double factor = 1;
+            double totalPp = 0;
+            double totalAccuracy = 0;
+
+            foreach (var item in filteredItems)
+            {
+                totalPp += item.Pp * factor;
+                totalAccuracy += item.Accuracy * factor;
+                factor *= 0.95;
```
roansong, line 60: So this calculates for every score of a user? I wonder if it makes sense to limit to just a user's top 100 or 200, given the diminishing returns. The "weird factor" below would need to then be pulled from a separate count of a user's total unique scores.

peppy: We're not about to be changing this. This is just following existing implementation which is tried and tested for years (see https://github.com/ppy/osu-performance).

smoogipoo:
Scores allowed to contribute to total pp/accuracy:
- Where `build_id = null`
- Where the given build has `allow_performance = 1`.

```sql
ALTER TABLE osu_builds
    ADD COLUMN `allow_performance` tinyint(1) NOT NULL DEFAULT '1'
    AFTER `allow_bancho`;
```

peppy:
This PR is fine to close right? Already being handled by #44?

smoogipoo:
Well, more like https://github.com/ppy/osu-queue-score-statistics/pull/45, but yeah it's fine to close.

PR #10, opened 2021-10-07
Use osu-beatmap-difficulty-cache to calculate difficulty attributes

smoogipoo:
- [x] Depends on https://github.com/ppy/osu-queue-score-statistics/pull/9

PRing mostly for comment for the time being. Perhaps we can try this out to see if the processor keeps up with the load?

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/AppSettings.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/AppSettings.cs
@@ -1,0 +1,20 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+
+namespace osu.Server.Queues.ScoreStatisticsProcessor
+{
+    public static class AppSettings
+    {
+        /// <summary>
+        /// The endpoint to query beatmap difficulties from.
+        /// </summary>
+        public static readonly string DIFFCALC_ENDPOINT;
+
+        static AppSettings()
+        {
+            DIFFCALC_ENDPOINT = Environment.GetEnvironmentVariable("DIFFCALC_ENDPOINT") ?? "http://localhost:5001/attributes";
+        }
+    }
+}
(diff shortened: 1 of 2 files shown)


osu.Server.Queues.ScoreStatisticsProcessor/Processors/PerformanceProcessor.cs
```diff
@@ -111,6 +105,55 @@ public static bool AllModsValidForPerformance(Mod[] mods)
...
+                BeatmapId = score.beatmap_id,
+                RulesetId = score.ruleset_id,
+                Mods = score.mods
+            }));
+
+            DifficultyAttributes? attributes = null;
+
+            req.Finished += () =>
+            {
+                if (req.ResponseStream == null)
```
peppy, line 129: Any reason to not just get the string immediately?

```suggestion
                if (!(req.GetResponseString() is string responseJson))
```

to avoid the null suppression below?

osu.Server.Queues.ScoreStatisticsProcessor/AppSettings.cs
```diff
@@ -0,0 +1,20 @@
...
+    public static class AppSettings
+    {
+        /// <summary>
+        /// The endpoint to query beatmap difficulties from.
+        /// </summary>
+        public static readonly string DIFFCALC_ENDPOINT;
+
+        static AppSettings()
+        {
+            DIFFCALC_ENDPOINT = Environment.GetEnvironmentVariable("DIFFCALC_ENDPOINT") ?? "http://localhost:5001/attributes";
```
peppy, line 17: I think this should probably be specified without the `attributes` route, so that if we ever need other exposed routes (quite possible) we don't need to change the envvar.

peppy (changes requested):

peppy:
I have the changes proposed ready to push if you're okay with them.

smoogipoo:
Changes sound reasonable.

peppy:
Won't merge this just yet, want to get stable out and see how things go there first.

smoogipoo:
Closing this for the time being. Can consider it again in the future.

PR #15, opened 2021-11-15
Replace DifficultyAttributeExtensions with game-side implementation

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu/pull/15633


osu.Server.Queues.ScoreStatisticsProcessor/Processors/PerformanceProcessor.cs
```diff
@@ -66,9 +71,40 @@ private double computePerformance(Ruleset ruleset, Mod[] mods, ScoreInfo score,
...
+            {
+                case 0:
+                    // Some attributes aren't databased because they're present in other tables.
+                    attributes = new OsuDifficultyAttributes
+                    {
+                        DrainRate = beatmap.diff_drain,
+                        HitCircleCount = beatmap.countNormal,
+                        SliderCount = beatmap.countSlider,
+                        SpinnerCount = beatmap.countSpinner
+                    };
```
smoogipoo, line 86: This is a little unfortunate. I'm all for making another class to hold these to be passed into `FromDifficultyAttributes()`, but please suggest a class name... `ExtraBeatmapDifficultyAttributes`?

PR #16, opened 2021-11-17
Add command to import osu_scores_high tables into solo_scores

smoogipoo:
- [x] https://github.com/ppy/osu/pull/15657

This is highly WIP, and shouldn't be merged before we discuss how to deal with the PKs and all.

It can be used for testing other score statistics processors as it is.

### DIFF
--- a/osu.Server.Queues.ScorePump/ImportHighScores.cs
+++ b/osu.Server.Queues.ScorePump/ImportHighScores.cs
@@ -91,9 +91,19 @@ namespace osu.Server.Queues.ScorePump
                     };
 
                     // Todo: Import highScore.hidden somehow?
-                    // Todo: Should we import pp into solo_scores_performance?
 
-                    db.Insert(soloScore);
+                    using (var transaction = db.BeginTransaction())
+                    {
+                        soloScore.id = db.Insert(soloScore, transaction);
+
+                        db.Execute("INSERT INTO solo_scores_performance (score_id, pp) VALUES (@ScoreId, @Pp)", new
+                        {
+                            ScoreId = soloScore.id,
+                            Pp = highScore.pp
+                        }, transaction);
+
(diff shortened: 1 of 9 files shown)


smoogipoo:
What `build_id` should scores from stable be imported with? Currently it imports with `null`. Will there be a special build providing a `build_id`?

peppy:
There are already `build_id`s for every stable build, not sure what to do about those that don't have that info though.

smoogipoo:
Yeah the question is which one to pick.

peppy:
I'd leave null if undetermined.

peppy:
@smoogipoo does this still need to be blocked? i think we can get it merged and work from there?

osu.Server.Queues.ScorePump/ImportHighScores.cs
```diff
@@ -0,0 +1,138 @@
...
+                            created_at = highScore.date,
+                            updated_at = highScore.date
+                        },
+                        created_at = highScore.date,
+                        updated_at = highScore.date,
+                    };
+
+                    // Todo: Import highScore.hidden somehow?
+
+                    using (var transaction = db.BeginTransaction())
```
peppy, line 95: Definitely don't want one transaction per insert, but can fix that when we get closer to running this on mass scores.

peppy (approved):

smoogipoo:
Doesn't need to be blocked, no, if you're fine with it as is.

osu.Server.Queues.ScorePump/ImportHighScores.cs
smoogipoo: One thing to note is that currently the previous score id is stored inside the json data. So right now `score.score_id != score.data.score_id` for these particular scores.

PR #19, opened 2021-11-29
Update imported highscore data ID to match primary key

smoogipoo:


osu.Server.Queues.ScorePump/ImportHighScores.cs
```diff
@@ -97,6 +97,9 @@ public int OnExecute(CancellationToken cancellationToken)
                     {
                         soloScore.id = db.Insert(soloScore, transaction);
 
+                        // Update data to match the row ID.
+                        db.Execute("UPDATE solo_scores s SET s.data = JSON_SET(s.data, '$.id', s.id) WHERE s.id = @id", soloScore, transaction);
```
peppy, line 101:
```suggestion
                        db.Execute("UPDATE solo_scores s SET s.data = JSON_SET(s.data, '$.id', @id) WHERE s.id = @id", soloScore, transaction);
```

I'd feel more comfortable with this if it works.

peppy (changes requested):

osu.Server.Queues.ScorePump/ImportHighScores.cs
smoogipoo: Works.

peppy (approved):

PR #25, opened 2022-02-01
Add support to `ImportHighScores` to populate old-new mapping pairs

peppy:
Not yet tested, PRing for discussion. Not 100% on the naming.


smoogipoo (approved):

smoogipoo:
Would probably prefer `scoreId` to be named `oldScoreId` but it's a bit of a much of a muchness I guess.

PR #26, opened 2022-02-02
Add play time processor

peppy:
The basics just to get the ball rolling.

We may want to start sending more accurate fail time data from lazer (ie. actual point in gameplay of failure, either as a 0..1 range or as a raw millisecond value based on gameplay time).

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Processors/PlayTimeProcessor.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Processors/PlayTimeProcessor.cs
@@ -40,15 +40,19 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Processors
             Debug.Assert(score.ended_at != null);
 
             // to ensure sanity, first get the maximum time feasible from the beatmap's length
-            int totalLengthSeconds = conn.QueryFirstOrDefault("SELECT total_length FROM osu_beatmaps WHERE beatmap_id = @beatmap_id", score);
+            double totalLengthSeconds = conn.QueryFirstOrDefault("SELECT total_length FROM osu_beatmaps WHERE beatmap_id = @beatmap_id", score);
 
-            // TODO: apply rate from mods.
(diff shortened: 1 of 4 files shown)


osu.Server.Queues.ScoreStatisticsProcessor/Processors/PlayTimeProcessor.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            Debug.Assert(score.started_at != null);
+            Debug.Assert(score.ended_at != null);
+
+            // to ensure sanity, first get the maximum time feasible from the beatmap's length
+            double totalLengthSeconds = conn.QueryFirstOrDefault<double>("SELECT total_length FROM osu_beatmaps WHERE beatmap_id = @beatmap_id", score, transaction);
+
+            Ruleset ruleset = ScoreStatisticsProcessor.AVAILABLE_RULESETS.Single(r => r.RulesetInfo.ID == score.ruleset_id);
+
+            var rateAdjustMods = score.mods.Select(m => m.ToMod(ruleset)).OfType<ModRateAdjust>().ToArray();
```
smoogipoo, line 50: I was going to suggest using `IApplicableToRate` because the whole `ModTimeRamp` subtree is excluded here, but that can be done later because it's a bit of a weird one.  
It would mean that you have to decide on a specific time to take the rate at and I'm not able to mentally prove if taking `base_duration / 2` as that time gives you the correct rate to approximate length in this case.

This is a fallback case if I understand it anyway so at worst it will be overestimating length by a bit.

smoogipoo (approved):
lgtm

osu.Server.Queues.ScoreStatisticsProcessor/Processors/PlayTimeProcessor.cs
peppy: Yeah, in theory I would hope we can trust start/end time due to other means of verification.

PR #28, opened 2022-02-24
Add automatic throttling to `ImportHighScores` based on replication slave latency

peppy:
Safari 2022-02-24 at 11 22 07

worked well until proxysql went and fucking crashed

### DIFF
--- a/osu.Server.Queues.ScorePump/ImportHighScores.cs
+++ b/osu.Server.Queues.ScorePump/ImportHighScores.cs
@@ -38,6 +38,7 @@ namespace osu.Server.Queues.ScorePump
         public long StartId { get; set; }
 
         private long lastCommitTimestamp;
+        private long lastLatencyCheckTimestamp;
 
         private static int currentReportInsertCount;
         private static int totalInsertCount;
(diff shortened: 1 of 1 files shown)


osu.Server.Queues.ScorePump/ImportHighScores.cs
```diff
@@ -125,9 +148,47 @@ void queueNextBatch()
...
+                Console.WriteLine($"Current slave latency of {latency} seconds exceeded maximum of {maximum_slave_latency_seconds} seconds.");
+                Console.WriteLine($"Sleeping for {latency} seconds to allow catch-up.");
+
+                Thread.Sleep(Math.Max(seconds_between_latency_checks * 2, latency * 1000));
+
+                // greatly reduce processing rate to allow for recovery.
+                scoresPerQuery = Math.Max(safe_minimum_scores_per_query, scoresPerQuery / 2);
+            }
+
+            if (latency > 0)
```
smoogipoo, line 180: So basically the only case where processing rate is allowed to be increased is when the slave latency is 0 seconds?

peppy: yep. it's usually always the case that it's 0 (or low enough that it floors to zero).

smoogipoo (approved):

PR #37, opened 2022-06-28
Fix mod masking differences from osu-performance

smoogipoo:
Two issues fixed with this one:
1. The old `LegacyModsHelper.keyMods` value doubled up `Key1`, but intended to use `KeyCoop` for that last one.
2. It didn't include all the new FL/FL+HD combos.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/LegacyModsHelper.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/LegacyModsHelper.cs
@@ -16,27 +16,20 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor
         // See: https://github.com/ppy/osu-performance/blob/83c02f50315a4ef7feea80acb84c66ee437d7210/include/pp/Common.h#L109-L129
         public static LegacyMods MaskRelevantMods(LegacyMods mods, bool isConvertedBeatmap, int rulesetId)
         {
-            LegacyMods relevantMods;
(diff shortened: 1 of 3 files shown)


osu.Server.Queues.ScoreStatisticsProcessor/LegacyModsHelper.cs
```diff
@@ -1,23 +1,49 @@
...
+        public static LegacyMods MaskRelevantMods(LegacyMods mods, bool isConvertedBeatmap, int rulesetId)
         {
-            LegacyMods relevantMods = LegacyMods.DoubleTime | LegacyMods.HalfTime | LegacyMods.HardRock | LegacyMods.Easy;
+            LegacyMods relevantMods;
+
+            switch (rulesetId)
+            {
+                case 0:
+                    if ((mods & LegacyMods.Flashlight) > 0)
+                        relevantMods = LegacyMods.DoubleTime | LegacyMods.HalfTime | LegacyMods.HardRock | LegacyMods.Easy | LegacyMods.Flashlight | LegacyMods.Hidden;
```
peppy, line 25: Could probably be simplified by moving the base case out to `relevantMods` definition above, unless I'm missing something?

```csharp
LegacyMods relevantMods = LegacyMods.DoubleTime | LegacyMods.HalfTime | LegacyMods.HardRock | LegacyMods.Easy;
```

smoogipoo: I suppose... I wanted to keep it matching osu-performance as close as possible.

peppy (approved):

PR #41, opened 2022-07-05
Update packages

smoogipoo:
Updates nuget packages + serves as a replacement for https://github.com/ppy/osu-queue-score-statistics/pull/15

Have tested to work.


osu.Server.Queues.ScoreStatisticsProcessor/LegacyRulesetHelper.cs
```diff
@@ -32,5 +37,26 @@ public static Ruleset GetRulesetFromLegacyId(int legacyId)
...
+                case 2:
+                    return new CatchDifficultyAttributes();
+
+                case 3:
+                    return new ManiaDifficultyAttributes();
+
+                default:
+                    throw new ArgumentException($"Invalid ruleset ID: {legacyId}", nameof(legacyId));
+            }
+        }
```
smoogipoo, line 60: Probably should be a method on `Ruleset` at some point, but this is probably fine for now.

peppy (approved):

Issue #40, opened 2022-07-05
High score importer doesn't set `started_at`


smoogipoo:
Pumping scores imported via the high score importer fails with:
```
Error processing score_id: 32479 user_id: 161270: System.AggregateException: One or more errors occurred. (Nullable object must have a value.)
 ---> System.InvalidOperationException: Nullable object must have a value.
   at System.Nullable`1.get_Value()
   at osu.Server.Queues.ScoreStatisticsProcessor.Processors.PlayCountProcessor.adjustMonthlyPlaycount(SoloScoreInfo score, MySqlConnection conn, MySqlTransaction transaction, Boolean revert) in /home/smgi/Repos/osu-queue-score-statistics/osu.Server.Queues.ScoreStatisticsProcessor/Processors/PlayCountProcessor.cs:line 53
   at osu.Server.Queues.ScoreStatisticsProcessor.Processors.PlayCountProcessor.ApplyToUserStats(SoloScoreInfo score, UserStats userStats, MySqlConnection conn, MySqlTransaction transaction) in /home/smgi/Repos/osu-queue-score-statistics/osu.Server.Queues.ScoreStatisticsProcessor/Processors/PlayCountProcessor.cs:line 32
   at osu.Server.Queues.ScoreStatisticsProcessor.ScoreStatisticsProcessor.ProcessResult(ScoreItem item) in /home/smgi/Repos/osu-queue-score-statistics/osu.Server.Queues.ScoreStatisticsProcessor/ScoreStatisticsProcessor.cs:line 86
   at osu.Server.QueueProcessor.QueueProcessor`1.ProcessResults(IEnumerable`1 items)
   at osu.Server.QueueProcessor.QueueProcessor`1.<>c__DisplayClass21_0.<Run>b__1()
   at System.Threading.Tasks.Task.InnerInvoke()
   at System.Threading.Tasks.Task.<>c.<.cctor>b__272_0(Object obj)
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location ---
   at System.Threading.Tasks.Task.ExecuteWithThreadLocal(Task& currentTaskSlot, Thread threadPoolThread)
   --- End of inner exception stack trace ---
```
Because the `PlayCountProcessor` and `PlayTimeProcessor` expect `started_at` and/or `ended_at` to be non-null:
[...]

PR #43, opened 2022-07-05
Import highscore date into `score->data->ended_at`

smoogipoo:
Fixes https://github.com/ppy/osu-queue-score-statistics/issues/40

I'm not sure how you'd go about fixing the already imported scores server-side - fastest for me was reimporting all scores.

Have tested to make sure the score processors are able to run.


peppy:
Already-imported are only there for testing purposes, and don't carry some other fixes anyways. Will be a full reimport at least once more.

peppy (approved):

peppy (approved):

Issue #36, opened 2022-06-27
Commands for processing all users' PP


smoogipoo:
Should match osu-performance as close as possible, bypassing the queue.

Issue #11, opened 2021-10-08
Support performance calculator blacklist


smoogipoo:
Probably should read from the existing database tables:
```sql
INSERT INTO `osu_beatmap_performance_blacklist`
VALUES (24722,0),(27638,0),(27737,0),(29844,0),(35015,0),(48874,0),(58064,0),(87570,0);
```

PR #44, opened 2022-07-05
Add commands to process score/total PP

smoogipoo:
Closes https://github.com/ppy/osu-queue-score-statistics/issues/34 (tested osu/taiko/catch/mania - no scores > 1pp difference, no totals >1pp difference).
Closes https://github.com/ppy/osu-queue-score-statistics/issues/35
Closes https://github.com/ppy/osu-queue-score-statistics/issues/36
Closes https://github.com/ppy/osu-queue-score-statistics/pull/6 (will be reimplemented)
Closes https://github.com/ppy/osu-queue-score-statistics/issues/11

This adds the following commands:
```
$ ./osu.Server.Queues.ScorePump

-> update-pp
    -> values
        -> all           | Computes raw PP values of all users.
        -> scores        | Computes raw PP values of select scores.
        -> sql           | Computes raw PP values of users given by an SQL statement.
        -> users         | Computes raw PP values of select users.
    -> totals
        -> all           | Computes total PP values of all users.
        -> users         | Computes total PP values select users.
```

The commands are made to match `osu-performance` with the exclusion of the ruleset argument:
```
osu-performance:      `-m [osu|taiko|catch|mania]`
osu-score-statistics: `-r [0|1|2|3]`
```

---
I've tested that PP values (scores and totals) are within 1pp difference to live values.

---

Scores:
- Does not process user totals.
- Considers the blacklist.
- Consider the valid ranking criteria (Ranked <= status <= Approved).
- Queries beatmaps/difficulty attributes from the database inline and caches progressively.
  - _Does not_ requery for beatmaps. `osu-performance` had a special case here but I assume it was because it loaded beatmaps at the start and had to requery beatmaps for newer scores. I deem this unnecessary.

Totals:
- Considers the blacklist.
- Stable builds are never excluded.
- Lazer builds (non-null `build_id`) are excluded unless allowed via `osu_builds.allow_performance`.
  - Currently this column doesn't exist so this is always `false`.
- Only calculates if the user has an existing `rank_score`.
- Does not zero out `rank_score` and `accuracy_new`.

---

Future work:
- Use this class in the statistics processor (I already have stashed changes).


osu.Server.Queues.ScoreStatisticsProcessor/DatabaseHelper.cs
```diff
@@ -0,0 +1,205 @@
...
+    {
+        /// <summary>
+        /// Retrieve user stats for a user based on a score context.
+        /// Creates a new entry if the user does not yet have one.
+        /// </summary>
+        /// <param name="score">The score to use for the user and ruleset lookup.</param>
+        /// <param name="db">The database connection.</param>
+        /// <param name="transaction">The database transaction, if one exists.</param>
+        /// <returns>The retrieved user stats. Null if the ruleset or user was not valid.</returns>
+        public static UserStats? GetUserStats(SoloScoreInfo score, MySqlConnection db, MySqlTransaction? transaction = null)
```
peppy, line 23: Do we really need this dual pathing? I'd prefer to async everything as discussed.

osu.Server.Queues.ScorePump/Performance/PerformanceMain.cs
```diff
@@ -0,0 +1,21 @@
...
+using McMaster.Extensions.CommandLineUtils;
+using osu.Server.Queues.ScorePump.Performance.Totals;
+using osu.Server.Queues.ScorePump.Performance.Values;
+
+namespace osu.Server.Queues.ScorePump.Performance
+{
+    [Command(Name = "update-pp", Description = "Computes the performance (pp) of scores.")]
+    [Subcommand(typeof(UpdateValuesMain))]
+    [Subcommand(typeof(UpdateTotalsMain))]
+    public sealed class PerformanceMain
```
peppy, line 13: `UpdatePerformanceCommand`?

osu.Server.Queues.ScorePump/Performance/PerformanceProcessor.cs
```diff
@@ -0,0 +1,345 @@
...
+
+namespace osu.Server.Queues.ScorePump.Performance
+{
+    /// <summary>
+    /// A helper class which can be used to compute scores' raw PP values and users' total PP values.
+    /// </summary>
+    public class PerformanceProcessor
+    {
+        private const BeatmapOnlineStatus min_ranked_status = BeatmapOnlineStatus.Ranked;
+        private const BeatmapOnlineStatus max_ranked_status = BeatmapOnlineStatus.Approved;
```
peppy, line 31: I'd prefer this was a static switch method rather than a range.

peppy:
JetBrains Rider-EAP 2022-07-06 at 04 42 17

osu.Server.Queues.ScorePump/Performance/PerformanceProcessor.cs
```diff
@@ -0,0 +1,345 @@
...
+
+        private readonly ConcurrentDictionary<int, Beatmap?> beatmapCache = new ConcurrentDictionary<int, Beatmap?>();
+        private readonly ConcurrentDictionary<DifficultyAttributeKey, BeatmapDifficultyAttribute[]?> attributeCache = new ConcurrentDictionary<DifficultyAttributeKey, BeatmapDifficultyAttribute[]?>();
+
+        private readonly ConcurrentDictionary<int, bool> builds;
+        private readonly ConcurrentDictionary<BlacklistEntry, byte> blacklist;
```
peppy, line 37: Should not be concurrent.

smoogipoo:
Yeah, I know... In https://github.com/ppy/osu-queue-score-statistics/pull/45 I've renamed the score statistics processor one to `ScorePerformanceProcessor`.

Guidance on naming would be appreciated, I'm coming up with blanks...

peppy (changes requested):
Initial pass, by no means comprehensive

peppy:
> Does not requery for beatmaps. osu-performance had a special case here but I assume it was because it loaded beatmaps at the start and had to requery beatmaps for newer scores. I deem this unnecessary.

It handled things like ranked state changes, which don't seem to be supported here currently. Not sure on the best way to handle this, but might be a good idea to share the logic I end up on with the `MetadataHub` to handle ongoing changes rather than requery the full set.

It also benefitted *heavily* from doing an initial lookup of all beatmaps rather than doing it one-by-one as required (super slow due to the number of individual queries). I doubt the version here will be able to keep up with production throughput immediately after starting up, until the beatmap cache warms up.

osu.Server.Queues.ScorePump/Performance/PerformanceMain.cs
smoogipoo: I named it this way since there's `PerformanceCommand` too, which is the basis for all other commands. This is only an entrypoint... Would you instead prefer I nest (private) these classes but keep the naming?

peppy: `PerformanceCommands`? That's how I named such cases these in ASS (`PrilivegedCommands` etc.).

smoogipoo:
Yeah it wouldn't be a big deal to do the initial lookup, but I was told and agreed with keeping it simple for the time being. I can do so if you wish.

osu.Server.Queues.ScorePump/Performance/PerformanceMain.cs
smoogipoo: Sure, sounds good.

peppy:
I'd feel safer having it done that way (fewer changes from the deployed version are better for now). Probably better in a separate PR though. I'll review this as-is since it's functional.

osu.Server.Queues.ScoreStatisticsProcessor/DatabaseHelper.cs
peppy: Hmm, this changed to be non-async only. Was there a reason for choosing that?

smoogipoo: I thought doing it this way basically serves as an extraction of the existing methods from `StatisticsProcessor`. I can use the async forms instead if you'd rather.

peppy: Just thinking that's the direction we want to go, if any.

osu.Server.Queues.ScorePump/Performance/Values/UpdateValuesScoresCommand.cs
```diff
@@ -0,0 +1,37 @@
...
+using McMaster.Extensions.CommandLineUtils;
+
+namespace osu.Server.Queues.ScorePump.Performance.Values
+{
+    [Command("scores", Description = "Computes pp of specific scores.")]
+    public class UpdateValuesScoresCommand : PerformanceCommand
+    {
+        [UsedImplicitly]
+        [Required]
+        [Argument(0, Description = "A space-separated list of scores to compute PP for.")]
```
peppy, line 18: Is this actually still space separated? I think McMaster is comma separated (which I'd prefer because this screws me over every time i run it on the old processor)

smoogipoo: Yeah it's still space separated. I'll use a string and do manual parsing.

peppy:
What would you say about renaming the commands to `UpdateScores` and `UpdateUserTotals`? Currently the naming is throwing me off (have to think "what are values and what are totals" each time).

smoogipoo:
Need more advice on naming of the individual classes.

```
UpdateUserTotalsCommands
UpdateScoresCommands

UpdateUserTotalsAllCommand
UpdateUserTotalsUsersCommand

UpdateScoresAllCommand
UpdateScoresScoresCommand
UpdateScoresSqlCommand
UpdateScoresUsersCommand
```
- Or renaming `UpdateUserTotalsUserCommand` -> `UpdateUserTotalsListCommand`, `UpdateScoresScoresCommand` -> `UpdateScoresListCommand`?
- Or removing the prefix and just having two `AllCommand`s in different namespaces?

peppy:
I think what you have in that list is fine enough. Might revisit later after working with them more.

peppy:
What I'm going to do is review this PR and apply naming / restructure in a separate PR. There's quite a bit I want to do there and it'll get messy here.

osu.Server.Queues.ScorePump/Performance/Totals/UpdateTotalsAllCommand.cs
```diff
@@ -0,0 +1,53 @@
...
+            int processedCount = 0;
+
+            await ProcessPartitioned(userIds, async id =>
+            {
+                using (var db = Queue.GetDatabaseConnection())
+                {
+                    var userStats = await DatabaseHelper.GetUserStatsAsync(id, RulesetId, db);
+
+                    // Only process users with an existing rank_score.
+                    if (userStats!.rank_score == 0)
```
peppy, line 40: use of `!` feels like abusive circumvention of the whole NRT pattern here..

osu.Server.Queues.ScorePump/Performance/Totals/UpdateTotalsAllCommand.cs
```diff
@@ -0,0 +1,53 @@
...
+
+            Console.WriteLine($"Processed 0 of {userIds.Length}");
+
+            int processedCount = 0;
+
+            await ProcessPartitioned(userIds, async id =>
+            {
+                using (var db = Queue.GetDatabaseConnection())
+                {
+                    var userStats = await DatabaseHelper.GetUserStatsAsync(id, RulesetId, db);
```
peppy, line 37: I really don't like having this same logic in two different files. I'll look to combine these.

osu.Server.Queues.ScorePump/Performance/Values/UpdateValuesAllCommand.cs
```diff
@@ -0,0 +1,92 @@
...
+                    break;
+
+                await ProcessPartitioned(users, async id =>
+                {
+                    using (var db = Queue.GetDatabaseConnection())
+                        await Processor.ProcessUserScoresAsync(id, RulesetId, db);
+                    Console.WriteLine($"Processed {Interlocked.Increment(ref processedCount)} of {totalCount}");
+                }, cancellationToken);
+
+                currentUserId = Math.Max(currentUserId, users.Max());
```
peppy, line 83: I'm trying to understand this `Math.Max`... is it here as a safety? Feels like it shouldn't need to be here in the first place.

smoogipoo: This code is mostly exactly mimicking what osu-performance does. I think you're right that it isn't required.

peppy:
I've made a seemingly final pass code-wise. @smoogipoo can you check to make sure you don't have any issues with the changes?

I'll do a second naming pass later on.

Note that I haven't tested actually running this extensively yet. Will do that over the coming days, using the staging setup.

peppy (approved):

osu.Server.Queues.ScorePump/Performance/PerformanceProcessor.cs
```diff
@@ -0,0 +1,374 @@
...
+        /// This does not insert the new stats values into the database.
+        /// </remarks>
+        /// <param name="rulesetId">The ruleset for which to update the total PP.</param>
+        /// <param name="userStats">An existing <see cref="UserStats"/> object to update with.</param>
+        /// <param name="connection">The <see cref="MySqlConnection"/>.</param>
+        /// <param name="transaction">An existing transaction.</param>
+        public async Task UpdateUserStatsAsync(UserStats userStats, int rulesetId, MySqlConnection connection, MySqlTransaction? transaction = null)
+        {
+            List<SoloScoreWithPerformance> scores = (await connection.QueryAsync<SoloScoreWithPerformance>(
+                $"SELECT `s`.*, `p`.`pp` AS `pp` FROM {SoloScore.TABLE_NAME} `s` "
```
peppy, line 265:
```suggestion
                $"SELECT `s`.*, `p`.`pp` FROM {SoloScore.TABLE_NAME} `s` "
```

Pretty sure the alias is unnecessary

osu.Server.Queues.ScorePump/Performance/PerformanceProcessor.cs
```diff
@@ -0,0 +1,374 @@
...
+        /// <param name="transaction">An existing transaction.</param>
+        public Task ProcessScoreAsync(SoloScore score, MySqlConnection connection, MySqlTransaction? transaction = null) => ProcessScoreAsync(score.ScoreInfo, connection, transaction);
+
+        /// <summary>
+        /// Processes the raw PP value of a given score.
+        /// </summary>
+        /// <param name="score">The score to process.</param>
+        /// <param name="connection">The <see cref="MySqlConnection"/>.</param>
+        /// <param name="transaction">An existing transaction.</param>
+        public async Task ProcessScoreAsync(SoloScoreInfo score, MySqlConnection connection, MySqlTransaction? transaction = null)
```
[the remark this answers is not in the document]
smoogipoo: Future use yes.

osu.Server.Queues.ScorePump/Performance/PerformanceProcessor.cs
```diff
@@ -0,0 +1,374 @@
...
+        /// Sets a count in the database.
+        /// </summary>
+        /// <param name="key">The count's key.</param>
+        /// <param name="value">The count's value.</param>
+        /// <param name="connection">The <see cref="MySqlConnection"/>.</param>
+        /// <param name="transaction">An existing transaction.</param>
+        public async Task SetCountAsync(string key, long value, MySqlConnection connection, MySqlTransaction? transaction = null)
+        {
+            await connection.ExecuteAsync("INSERT INTO `osu_counts` (`name`,`count`) VALUES (@Name, @Count) "
+                                          + "ON DUPLICATE KEY UPDATE `name` = VALUES(`name`), `count` = VALUES(`count`)", new
```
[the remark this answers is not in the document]
smoogipoo: Copied from osu-performance. Sure.

smoogipoo:
> can you check to make sure you don't have any issues with the changes?

Looks fine.

peppy:
Are you planning to make a migration for the `allow_performance` column over at osu-web or shall I do that?

smoogipoo:
Can do.

peppy (approved):

PR #45, opened 2022-07-05
Make pp score statistics processor use the new `PerformanceProcessor`

smoogipoo:
Prereqs:
- [ ] https://github.com/ppy/osu-queue-score-statistics/pull/44

Naming is hard, accepting suggestions. The statistics processor has been renamed to `ScorePerformanceProcessor`, and the new class has been kept as is.


smoogipoo:
Just as a note, merging this will cause user totals to be reprocessed by this project upon new scores arriving in the queue. We may want to turn it off until `osu-performance` is turned off?

smoogipoo:
I'm gonna close this one for the time being, it'll return after #44 is merged.

PR #47, opened 2022-07-08
Restructure commands and naming

peppy:
Probably easiest to review this in solution explorer and `dotnet run`.

I still feel like combining the commands which get users/scores from different sources into single commands with multiple optional filter types may work best in terms of making the command easier to use.

ie. I'd like to run `dotnet run batch scores --sql="select user_id ..."` and if providing no filter it would run on all.

Feels more conventional? not sure.


osu.Server.Queues.ScorePump/Performance/Scores/UpdateScoresForUsers.cs
```diff
@@ -7,10 +7,10 @@
 using JetBrains.Annotations;
 using McMaster.Extensions.CommandLineUtils;
 
-namespace osu.Server.Queues.ScorePump.Performance.Values
+namespace osu.Server.Queues.ScorePump.Performance.Scores
 {
-    [Command("users", Description = "Computes pp of specific users.")]
-    public class UpdateValuesUsersCommand : PerformanceCommand
+    [Command("userlist", Description = "Computes pp of all scores of specific users.")]
```
smoogipoo, line 12: Just confirming, you intended `userlist` instead of `users`?

peppy: Yeah... because it gets confusing otherwise like `dotnet run batch scores users` and you also have `dotnet run batch user-totals`

Would be resolved if the filter commands are combined as I touch on in OP. Can look at doing that if you agree with the direction.

smoogipoo: I'm like 60% in agreement only because I can foresee issues with what if multiple options are given. But at the end of the day you're the one going to be using these commands, so...

smoogipoo (approved):

osu.Server.Queues.ScorePump/Performance/Scores/UpdateScoresForUsers.cs
peppy: Will leave them for now and see how they feel.

PR #49, opened 2022-07-08
Move performance processor to score statistics project

smoogipoo:
I want to use this new `PerformanceProcessor` for the score statistics (watch command) too, so I'm moving it to the score statistics project in preparation for that.

The existing `PerformanceProcessor` in that project (handling incoming scores) is renamed to `ScorePerformanceProcessor`. Please give any naming suggestions.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor.Tests/PerformanceProcessorTests.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor.Tests/PerformanceProcessorTests.cs
@@ -68,7 +68,7 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Tests
             };
 
             foreach (var mod in mods)
-                Assert.True(PerformanceProcessor.AllModsValidForPerformance(new[] { mod }), mod.GetType().ReadableName());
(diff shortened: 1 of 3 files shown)


peppy:
I'm fine with `ScorePerformanceProcessor` (guessing there will be a `UserTotalPerformanceProcessor` alongside or similar?) but we really can't have `PerformanceProcessor`, since it's not a "processor" as defined by this project.

I also can't think of a better name for it. An alternative might be to rename all the existing processors to `Updater`s instead?

smoogipoo:
> guessing there will be a `UserTotalPerformanceProcessor` alongside or similar

Hmm, in my changes I originally batched it in the one processor to avoid ordering weirdness, but I suppose that can be done via an order property.

> An alternative might be to rename all the existing processors to `Updater`s instead

Or rename the new one to `PerformanceUpdater`?

peppy:
I guess `PerformanceUpdater` works for now. Is there any reason it does both the user totals and the per-pp updates in a single file btw? If that could be split out (ie. no shared logic) that would probably be best.

smoogipoo:
Gonna close this. I think https://github.com/ppy/osu-queue-score-statistics/pull/50 is a better path.

PR #50, opened 2022-07-08
Split PerformanceProcessor implementation to statistics processor classes

smoogipoo:
Alternative to / closes https://github.com/ppy/osu-queue-score-statistics/pull/49

Taking a bit of a different approach here following the suggestion in https://github.com/ppy/osu-queue-score-statistics/pull/49#issuecomment-1178854593.

- `PerformanceProcessor` has been split into the existing statistics processor classes, renaming one of them to `ScorePerformanceProcessor` in the process.

- `BeatmapStore` and `BuildStore` added to handle retrieval of the databased objects. The `osu_counts` IO methods have been moved to `DatabaseHelper`.

- `UserTotalPerformanceProcessor`'s "processing" part isn't actually implemented yet (future PR).
- `ScorePerformanceProcessor`'s "processing" part still does its own thing (https://github.com/ppy/osu-queue-score-statistics/pull/51).

- No functional changes.

PR #52, opened 2022-07-08
Add user total performance processing

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu-queue-score-statistics/pull/50

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserTotalPerformanceProcessor.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserTotalPerformanceProcessor.cs
@@ -32,15 +32,15 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Processors
         {
             var dbInfo = LegacyDatabaseHelper.GetRulesetSpecifics(score.ruleset_id);
 
-            UpdateUserStatsAsync(userStats, score.ruleset_id, conn, transaction).Wait();
-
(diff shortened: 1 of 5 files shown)


smoogipoo:
Just as a note, since this is going to be writing to the `osu_user_stats` tables, you'll probably want to create a dummy table for your test runs.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserTotalPerformanceProcessor.cs
```diff
@@ -21,12 +21,26 @@ public class UserTotalPerformanceProcessor : IProcessor
...
         {
         }
 
         public void ApplyToUserStats(SoloScoreInfo score, UserStats userStats, MySqlConnection conn, MySqlTransaction transaction)
         {
+            var dbInfo = LegacyDatabaseHelper.GetRulesetSpecifics(score.ruleset_id);
+
+            UpdateUserStatsAsync(userStats, score.ruleset_id, conn, transaction).Wait();
+
+            int warnings = conn.QuerySingleOrDefault<int>($"SELECT `user_warnings` FROM {dbInfo.UsersTable} WHERE `user_id` = @UserId", new
```
peppy, line 37: I don't understand this. Shouldn't it just skip the `UpdateUserStatsAsync` call rather than zero the value out? Is there a reason to firs calculate and then reset?

smoogipoo: Nope, I had it written this way in another branch and didn't double-take.

Issue #13, opened 2021-11-01
Use SubmittableScore from osu.Game for score data


smoogipoo:
Added in https://github.com/ppy/osu/pull/15351, which is the new concrete model for score submissions.

PR #57, opened 2022-07-12
Update to use shared model class from `osu.Game`

peppy:
Resolves https://github.com/ppy/osu-queue-score-statistics/issues/13

- [x] Depends on ppy/osu#19082.


peppy:
I haven't heavily tested this at runtime yet. Will do so over the coming days (will use it to re-run some imports at least).

smoogipoo:
Need to go through this with a fine comb. Things like this will fail (I believe): https://github.com/ppy/osu-queue-score-statistics/blob/f288bd7abbe8d29fa2fe07968b1d0ef0e21230f3/osu.Server.Queues.ScoreStatisticsProcessor/Processors/PlayTimeProcessor.cs#L45

It's also why I advocate to use
```
new
{
    X = score.X
}
```
At all times in our queries.

peppy:
yeah, it's a valid concern.

smoogipoo (approved):
I think this should be fine. Also haven't tested but can try to in the next ~2 hours.

smoogipoo (approved):
It seems to be okay.

PR #60, opened 2022-07-15
Push scores to ES indexer queue after processing is complete

peppy:
Not yet tested, but should be fine.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/ScoreStatisticsProcessor.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/ScoreStatisticsProcessor.cs
@@ -30,6 +30,8 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor
 
         private readonly List<IProcessor> processors = new List<IProcessor>();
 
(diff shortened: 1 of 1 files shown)


peppy:
Hold off on this, apaprently the model is not the same format via *attributes on the class* 🤦 

https://github.com/ppy/osu-elastic-indexer/blob/be950c70dfde31bfcbec108061ad5aa74ffc30c0/osu.ElasticIndexer/SoloScore.cs#L17-L23

osu.Server.Queues.ScoreStatisticsProcessor/ScoreStatisticsProcessor.cs
```diff
@@ -137,5 +144,26 @@ private static List<Ruleset> getRulesets()
 
             return rulesetsToProcess;
         }
+
+        private class ElasticQueueProcessor : QueueProcessor<ElasticQueueProcessor.ElasticScoreItem>
```
peppy, line 148: Maybe we can pull this in via nuget in the future, dunno if it's worth it.

peppy:
This is deployed and working (https://github.com/ppy/osu-queue-score-statistics/compare/production?expand=1). Should be fine to review now.

smoogipoo (approved):

PR #62, opened 2022-07-22
Mark passed scores as preserve

peppy:
Eventually this will be handled by osu-web I think. Already tested on production(tm).

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/ScoreStatisticsProcessor.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/ScoreStatisticsProcessor.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Reflection;
+using Dapper;
 using Dapper.Contrib.Extensions;
 using MySqlConnector;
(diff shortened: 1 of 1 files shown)


smoogipoo:
I suppose it make more sense to be here considering pp maybe?

smoogipoo:
Can this be made into its separate processor? Dunno when/how this flag is going to be used, but I for example see the flag being determined by whether the score is a new user best PP, and wouldn't want to add more logic to this entry class.

peppy:
If it gets more complex then this then yeah, for sure.

My general plan is to have a completely separate component manage the `preserve` flag, rather than do it here or osu-web. This is just a sane default to handle the generic case.

I foresee this being moved to `osu-web` in the near future, just wanted to get this deployable for now.

peppy:
To elaborate: it doesn't matter if more than one `user-beatmap(-mod)` combinations are `preserve`d. osu-web already has deduping at its end to ensure only a single distinct score is returned per user as expected.

So the flag is mostly for optimisation and cleanup. Err in the direction of having it `true` if unsure.

smoogipoo (approved):

PR #75, opened 2022-08-25
Backport: Add osu!lazer score `maximum_statistics` upgrade command

smoogipoo:
**This is a `release` branch pull request!!**

Also contains https://github.com/ppy/osu-queue-score-statistics/pull/74  
This has only been PR'd to this branch because it's expected to be executed only once.

The `add-maximum-statistics-to-lazer-scores` command adds the `maximum_statistics` dictionary to all scores in the database.

I previously asked for the last 10000 scores in the `solo_scores` table. Here's the results of running this command on them: https://drive.google.com/file/d/1cbcTZEBtU1GnM4-unCZF5g8LP9or4y4U/view?usp=sharing **This is a 10MB text file, be careful!!**

### DIFF
--- a/osu.Server.Queues.ScorePump/AddMaximumStatisticsToLazerScores.cs
+++ b/osu.Server.Queues.ScorePump/AddMaximumStatisticsToLazerScores.cs
@@ -50,10 +50,7 @@ namespace osu.Server.Queues.ScorePump
                         }, transaction)).ToArray();
 
                         if (scores.Length == 0)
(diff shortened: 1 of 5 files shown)


smoogipoo:
Also, we should probably ensure no new scores can arrive into the `solo_scores` table without `maximum_statistics`. Re-running the script won't have any adverse effects though.

peppy (approved):

PR #76, opened 2022-08-26
Make the highscore importer provide the `MaximumStatistics` dictionary

smoogipoo:
Prereqs:
- [x] Updated osu.Game package.
- [x] osu-difficulty-calculator to have been re-run for osu!mania with https://github.com/ppy/osu-difficulty-calculator/tree/release-diffcalc + https://github.com/ppy/osu/tree/release-diffcalc
- [x] https://github.com/ppy/osu/pull/20028

The purpose of this PR is to add the `MaximumStatistics` dictionary to imported scores.
- First we convert every hit type to the corresponding "accurate" version. For example for an `ok` result, the accurate version is `great` in osu/taiko/catch and `perfect` in mania. This is done via [`Ruleset.GetHitResults()`](https://github.com/ppy/osu/blob/289b6f1a583dc50e84cadff713808f43fddc7ff0/osu.Game/Rulesets/Ruleset.cs#L310-L339) as mentioned at in my [original PR](https://github.com/ppy/osu/pull/19939).
- Then, we compare the combo from those statistics to the combo in the `osu_beatmap_difficulty_attributes` table (`attrib_id = 9`), and add a `legacy_combo_increase` hit result to bring them up to the same level if they don't match. This only shows up for osu/mania - taiko/catch have accurate combo counts from score statistics alone.

I have checked that combo count is accurate for osu, **however** mania combo is a few combo off because osu!stable is silly and allows players to get more maximum combo than should actually be possible via hold notes. I'm unsure how to deal with this yet - but I gather we can do a separate pass over it later on.

Additionally, this PR migrates total score into `LegacyTotalScore`.

Sample files:
[osu.txt](https://github.com/ppy/osu-queue-score-statistics/files/9450699/osu.txt)
[taiko.txt](https://github.com/ppy/osu-queue-score-statistics/files/9450701/taiko.txt)
[catch.txt](https://github.com/ppy/osu-queue-score-statistics/files/9450704/catch.txt)
[mania.txt](https://github.com/ppy/osu-queue-score-statistics/files/9450706/mania.txt)

Imported from the top-1000 dumps on https://data.ppy.sh

### DIFF
--- a/osu.Server.Queues.ScorePump/Queue/ImportHighScores.cs
+++ b/osu.Server.Queues.ScorePump/Queue/ImportHighScores.cs
@@ -284,14 +284,15 @@ namespace osu.Server.Queues.ScorePump.Queue
                             BeatmapID = highScore.beatmap_id,
                             RulesetID = ruleset.RulesetInfo.OnlineID,
                             Passed = true,
-                            TotalScore = highScore.score,
+                            TotalScore = (int)referenceScore.TotalScore,
                             Accuracy = referenceScore.Accuracy,
                             MaxCombo = highScore.maxcombo,
                             Rank = Enum.TryParse(highScore.rank, out ScoreRank parsed) ? parsed : ScoreRank.D,
                             Mods = referenceScore.Mods.Select(m => new APIMod(m)).ToArray(),
                             Statistics = referenceScore.Statistics,
                             MaximumStatistics = referenceScore.MaximumStatistics,
-                            EndedAt = highScore.date
+                            EndedAt = highScore.date,
+                            LegacyTotalScore = highScore.score
                         });
 
                         insertCommand.Transaction = transaction;
@@ -412,6 +413,13 @@ namespace osu.Server.Queues.ScorePump.Queue
                 if (maxComboFromAttributes > maxComboFromStatistics)
                     scoreInfo.MaximumStatistics[HitResult.LegacyComboIncrease] = maxComboFromAttributes.Value - maxComboFromStatistics;
 #pragma warning restore CS0618
+
+                ScoreProcessor scoreProcessor = ruleset.CreateScoreProcessor();
+                int baseScore = scoreInfo.Statistics.Where(kvp => kvp.Key.AffectsAccuracy()).Sum(kvp => kvp.Value * Judgement.ToNumericResult(kvp.Key));
+                int maxBaseScore = scoreInfo.MaximumStatistics.Where(kvp => kvp.Key.AffectsAccuracy()).Sum(kvp => kvp.Value * Judgement.ToNumericResult(kvp.Key));
+
+                scoreInfo.TotalScore = (int)scoreProcessor.ComputeScore(ScoringMode.Standardised, scoreInfo);
+                scoreInfo.Accuracy = maxBaseScore == 0 ? 1 : baseScore / (double)maxBaseScore;
 
                 return scoreInfo;
             }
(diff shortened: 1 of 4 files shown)


smoogipoo:
I've bundled the LegacyTotalScore migration here since it's kinda related to everything anyway.

osu.Server.Queues.ScorePump/Queue/ImportHighScores.cs
```diff
@@ -284,14 +284,15 @@ public async Task Run(HighScore[] scores)
                             BeatmapID = highScore.beatmap_id,
                             RulesetID = ruleset.RulesetInfo.OnlineID,
                             Passed = true,
-                            TotalScore = highScore.score,
+                            TotalScore = (int)referenceScore.TotalScore,
```
peppy, line 287: scary cast

peppy:
I'm very concerned about the performance implications of these changes (+1RTT per score, taking the count up from zero). Will do a test run tonight and see how things go.

No need to make changes just yet, but I'm pretty sure we'll want to fetch every combo for every beatmap before running.

osu.Server.Queues.ScorePump/Queue/ImportHighScores.cs
```diff
@@ -413,6 +414,13 @@ private async Task<ScoreInfo> createReferenceScore(Ruleset ruleset, HighScore hi
                     scoreInfo.MaximumStatistics[HitResult.LegacyComboIncrease] = maxComboFromAttributes.Value - maxComboFromStatistics;
 #pragma warning restore CS0618
 
+                ScoreProcessor scoreProcessor = ruleset.CreateScoreProcessor();
+                int baseScore = scoreInfo.Statistics.Where(kvp => kvp.Key.AffectsAccuracy()).Sum(kvp => kvp.Value * Judgement.ToNumericResult(kvp.Key));
+                int maxBaseScore = scoreInfo.MaximumStatistics.Where(kvp => kvp.Key.AffectsAccuracy()).Sum(kvp => kvp.Value * Judgement.ToNumericResult(kvp.Key));
+
+                scoreInfo.TotalScore = (int)scoreProcessor.ComputeScore(ScoringMode.Standardised, scoreInfo);
+                scoreInfo.Accuracy = maxBaseScore == 0 ? 1 : baseScore / (double)maxBaseScore;
```
frenzibyte, line 422: I'm very likely to be missing something here, but does this apply to legacy/stable scores? And if so, wouldn't the accuracy change here because slider ticks/ends are not present in the statistics? This is with the assumption that slider ticks/ends affect accuracy in stable.

smoogipoo: > does this apply to legacy/stable scores?

This entire thing applies only to legacy/stable scores.

> wouldn't the accuracy change

Yes, the accuracy does change. However it's not exactly how you think of it. Slider ticks/ends do not affect accuracy in stable, and the values computed by both osu!stable and osu!web are derived solely from the statistics available in scores.

Some cases do differ, such as mania, where in stable/web GREATs/PERFECTs are treated as the same value, but in lazer they're not. For taiko, stable uses 300/150 whereas lazer uses 300/100.

This information is recoverable, because, as I said, it's done in [real time from score statistics](https://github.com/ppy/osu/blob/8c202ce141c3f4b742e401caf0b47bb694a528d5/osu.Game/Scoring/Legacy/LegacyScoreDecoder.cs#L138-L235).

frenzibyte: Right, thanks for the clarification.

peppy (approved):

PR #85, opened 2022-09-09
Use `ulong` for highscore imports

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu/pull/20202
- [ ] https://github.com/ppy/osu-web/pull/9264

---

- `ProcessHistory` left as-is because it's defined as signed in the database.
- `ElasticScoreItem` left as-is because I'm not sure if there'd be any effects in changing the data type yet. Can be done later but isn't really what I'm trying to achieve.


peppy:
Will just merge this ahead of the osu-web change, as I won't be deploying a new version before those changes are made (and well, most of them if not all are already applied in production).

peppy (approved):

PR #87, opened 2022-09-12
Add command to upgrade scores

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu/pull/20262

This PR ports a command which was pushed to the `release` branch, as well adds a new function to it to upgrade existing total score and accuracy values.

The command which was ported from the `release` branch is: https://github.com/ppy/osu-queue-score-statistics/blob/3b329c474e93f858ffad5c6b5d659ba115c7eff6/osu.Server.Queues.ScorePump/AddMaximumStatisticsToLazerScores.cs
This was done mainly because upgrading total score and accuracy values requires `MaximumStatistics` to be fully populated. And as we haven't yet blocked old lazer clients from submitting scores, some still have empty `MaximumStatistics`.

As for the total score upgrades... Of the 10000 scores that were provided to me, here's the full list of changes:
[changes.txt](https://github.com/ppy/osu-queue-score-statistics/files/9547141/changes.txt)
I've gone through some of these rows to identify the reason for the changes:
```
##
## This is a passed=false score which has statistics populated with the full values.
## This comes from a time where we used to mangle statistics with the yet-to-be-completed judgements.
## We no longer do this mangling as of https://github.com/ppy/osu/pull/20010.
##
Updated 48087761 (22515, 57.46%) -> (22515, 2.33%) - (+0.00%, -95.94%)

##
## This is a passed=false RX score? Not sure how this one is possible.
## The mod multiplier was reduced to 10% as of https://github.com/ppy/osu/pull/19715.
##
Updated 48087628 (97389, 97.45%) -> (9739, 10.75%) - (-90.00%, -88.96%)

##
## This is a WU score.
## The mod multiplier was reduced to 50% as of https://github.com/ppy/osu/pull/20206.
##
Updated 48086384 (173285, 71.39%) -> (86642, 71.39%) - (-50.00%, +0.00%)
```
Of those scores, unfortunately, it seems the first case is the most prevalent. Perhaps in the future we could try to use play length as a way to reduce the number of misses stored in these scores' statistics.

### DIFF
--- a/osu.Server.Queues.ScorePump/Queue/UpgradeScores.cs
+++ b/osu.Server.Queues.ScorePump/Queue/UpgradeScores.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
@@ -12,6 +13,7 @@ using McMaster.Extensions.CommandLineUtils;
 using osu.Game.Rulesets;
 using osu.Game.Rulesets.Judgements;
 using osu.Game.Rulesets.Scoring;
+using osu.Game.Scoring;
 using osu.Server.Queues.ScoreStatisticsProcessor;
 using osu.Server.Queues.ScoreStatisticsProcessor.Models;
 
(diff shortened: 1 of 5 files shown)


peppy:
The first case should not come up for future scores, correct?

smoogipoo:
Yeah. It can be removed once lazer scores are not able to be submitted from those old builds anymore.

peppy:
To confirm I should disable old build submission before doing a final run of this?

peppy (approved):

smoogipoo:
Yes, but it can also be run as many times as needed until then.

PR #95, opened 2022-10-05
Fix imported highscores not including classic mod

smoogipoo:


peppy:
Not sure what happened to github actions, will force push to attempt to recover it.

peppy:
🤷

PR #99, opened 2022-10-21
Limit precision of accuracy storage during high score import

peppy:


osu.Server.Queues.ScorePump/Queue/ImportHighScores.cs
```diff
@@ -309,7 +309,7 @@ public async Task Run(HighScore[] scores)
                             RulesetID = ruleset.RulesetInfo.OnlineID,
                             Passed = true,
                             TotalScore = (int)referenceScore.TotalScore,
-                            Accuracy = referenceScore.Accuracy,
+                            Accuracy = Math.Round(referenceScore.Accuracy, 4, MidpointRounding.ToEven),
```
frenzibyte, line 312: Same as https://github.com/ppy/osu/pull/20847#discussion_r1002085986 to some extent.

frenzibyte (changes requested):

peppy:
Going to close this for now. As I found out, this doesn't affect the cost of storing accuracy, so we can just leave it until next time the precision causes an issue for us.

PR #106, opened 2022-11-29
Update osu nuget packages

smoogipoo:
Note that lazer taiko scores' PP values will differ from what's expected by the deployed version of `osu-performance` due to https://github.com/ppy/osu/pull/20558, but this isn't a huge deal for now I think?

I'm just doing this to remove the rounding that breaks compilation.


smoogipoo:
This one might be broken. Seems like taiko ruleset failed to deploy last time: https://ci.appveyor.com/project/peppy/osu-ehtfo/build/job/jfh18sekt29kqor7

peppy (approved):

PR #107, opened 2022-11-29
Add medal processor

peppy:
PRing ahead of adding proper testability (currently looking at updating tests to user a more portable means).

Currently only processes a minimal subset of medals, but is designed to be modular and performant enough to handle all existing medals.

- [x] Depends on #110 
- [x] Depends on #114

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor.Tests/MedalProcessorTests.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor.Tests/MedalProcessorTests.cs
@@ -1,0 +1,72 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using Dapper.Contrib.Extensions;
+using MySqlConnector;
+using Xunit;
+
+namespace osu.Server.Queues.ScoreStatisticsProcessor.Tests
+{
+    public class MedalProcessorTests : DatabaseTest
+    {
+        [Fact]
+        public void TestSimplePackAwarding()
+        {
+            /*
+                // Pass all songs in Video Game Pack vol.1
+
+                | pack_id | beatmapset_id | min(beatmap_id) |
+                |---------|---------------|-----------------|
+                | 40      | 13022         | 71621           |
+                | 40      | 16520         | 59225           |
+                | 40      | 23073         | 79288           |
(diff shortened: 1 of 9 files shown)


peppy:
I'm still continuing forward work on this, but also believe this PR is in a good state to be reviewed and potentially merged.

osu.Server.Queues.ScoreStatisticsProcessor.Tests/MedalProcessorTests.cs
```diff
@@ -0,0 +1,198 @@
...
+                setScoreForBeatmap(beatmap.beatmap_id);
+            }
+
+            assertAwarded(medal_id);
+
+            WaitForDatabaseState("SELECT playcount FROM osu_user_stats WHERE user_id = 2", AllBeatmaps.Count, CancellationToken);
+        }
+
+        [Fact]
+        public void TestPlayMultipleBeatmapsFromSameSet()
```
smoogipoo, line 88: As discussed IRL, I think these names should be updated to reflect the assertion. For this case maybe something like `TestPlayMultipleBeatmapsFromSameSetDoesNotAward`.

smoogipoo (changes requested):

osu.Server.Queues.ScoreStatisticsProcessor.Tests/MedalProcessorTests.cs
peppy: I've added more documentation and tidied things up. See how it reads now.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/IMedalAwarder.cs
```diff
@@ -0,0 +1,28 @@
...
+using osu.Game.Online.API.Requests.Responses;
+using osu.Server.Queues.ScoreStatisticsProcessor.Models;
+
+namespace osu.Server.Queues.ScoreStatisticsProcessor.Processors
+{
+    /// <summary>
+    /// A class which handles conditional awarding of one of more medals.
+    /// </summary>
+    [UsedImplicitly]
+    internal interface IMedalAwarder
```
smoogipoo, line 16: Probably want this to be `public` for hush-hush.

```suggestion
    public interface IMedalAwarder
```

smoogipoo (approved):
Seems okay.

PR #118, opened 2022-12-05
Target alternative pp column

peppy:


peppy:
Will wait for upcoming osu-web migration for this, which will make the tests work again.

peppy:
This should be good to go!

smoogipoo (approved):

PR #119, opened 2022-12-06
Tidy up command structure and naming

peppy:


smoogipoo (approved):
I'd say the `import-high-scores` command could/should be at the higher level instead of under `queue` (so, on the same level as `recalculate-scores`), but let's go with this for now.

peppy:
I'll move it in my next PR which modifies that command.

peppy:
Well actually I'm not sure it should be moved... one of the pieces I'm looking to add to that command is to queue for indexing as part of the import. What do you think?

smoogipoo:
Hard to say. By that logic the recalculate-scores should also be moved in there because it also needs to requeue for indexing if it's not already.

peppy:
Well... you'll see why I want indexing in the next PR. I still wouldn't be adding it to `recalculate-scores` because that's intended to run over every score in existence.

PR #121, opened 2022-12-06
Add support for realtime legacy score importing (and automatic indexing)

peppy:
I was originally going to write this as a separate command... but it turns out that it's probably lower effort on all parts to reuse this existing command. Abstracting out the shared parts to a base class was a bit of a pain.

I've tested this locally and it seems to work quite well:

https://user-images.githubusercontent.com/191335/205884294-f76f02ba-2835-4511-9801-253d54bcd096.mp4

### DIFF
--- a/osu.Server.Queues.ScorePump/Queue/ImportHighScoresCommand.cs
+++ b/osu.Server.Queues.ScorePump/Queue/ImportHighScoresCommand.cs
@@ -37,16 +37,27 @@ namespace osu.Server.Queues.ScorePump.Queue
         public int RulesetId { get; set; }
 
         /// <summary>
-        /// The high score ID to start the import process from. This can be used to resume an existing job, or perform catch-up on new scores.
+        /// The high score ID to start the import process from. This can be used to perform batch reimporting for special cases.
         /// </summary>
         [Option(CommandOptionType.SingleValue)]
-        public ulong StartId { get; set; }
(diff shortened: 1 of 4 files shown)


osu.Server.Queues.ScorePump/Queue/ImportHighScoresCommand.cs
```diff
@@ -37,20 +37,39 @@ public class ImportHighScoresCommand : BaseCommand
...
+        /// <summary>
+        /// Whether to skip pushing imported score to the elasticsearch indexing queue.
+        /// </summary>
+        [Option(CommandOptionType.SingleValue, Template = "--skip-indexing")]
+        public bool SkipIndexing { get; set; }
+
+        /// <summary>
+        /// Whether to exit when there are no scores left at the tail end of the import. Defaults to <c>false</c>.
+        /// </summary>
+        public bool ExitOnCompletion { get; set; }
```
smoogipoo, line 66: Was this intended to have an `[Option]` on it?

smoogipoo (approved):
Looks fine otherwise.

osu.Server.Queues.ScorePump/Queue/ImportHighScoresCommand.cs
peppy: Yeah.

PR #122, opened 2022-12-09
Calculate `rank_index` as part of performance total process

peppy:


osu.Server.Queues.ScoreStatisticsProcessor.Tests/PerformanceProcessorTests.cs
```diff
@@ -1,19 +1,81 @@
...
+            }
+
+            SetScoreForBeatmap(TEST_BEATMAP_ID, score =>
+            {
+                score.Score.ScoreInfo.Statistics[HitResult.Great] = 100;
+                score.Score.ScoreInfo.MaxCombo = 100;
+                score.Score.ScoreInfo.Accuracy = 1;
+            });
+
+            WaitForDatabaseState("SELECT rank_score_exp FROM osu_user_stats WHERE user_id = 2", 1082, CancellationToken);
```
smoogipoo, line 75: Can it not be hardcoded? PP values change all the time and I don't want to update yet another test (along with the osu!-side ones) whenever a change is made. Just `> 0`? `!= 0`?

smoogipoo (changes requested):

osu.Server.Queues.ScoreStatisticsProcessor.Tests/PerformanceProcessorTests.cs
peppy: Have adjusted.

smoogipoo (approved):

PR #126, opened 2022-12-27
Fix max combo missed during solo-score import

smoogipoo:
Fixes https://github.com/ppy/osu-web/issues/9681

### DIFF
--- a/osu.Server.Queues.ScorePump/Queue/ImportHighScoresCommand.cs
+++ b/osu.Server.Queues.ScorePump/Queue/ImportHighScoresCommand.cs
@@ -451,6 +451,7 @@ namespace osu.Server.Queues.ScorePump.Queue
                     Mods = ruleset.ConvertFromLegacyMods((LegacyMods)highScore.enabled_mods).Append(classicMod).ToArray(),
                     Statistics = new Dictionary<HitResult, int>(),
                     MaximumStatistics = new Dictionary<HitResult, int>(),
+                    MaxCombo = highScore.maxcombo
                 };
 
                 // Populate statistics and accuracy.


peppy:
This is going to require another multi-month upgrade operation I guess? With an extra query to fetch from the legacy table?

peppy:
Can this be marked as closing https://github.com/ppy/osu-web/issues/9681?

smoogipoo:
Unfortunately, yes :(

smoogipoo:
Also, this only affects the import process. As far as I can tell, `recalculate-scores` is working as intended, so it's highly likely that the scores will be fixed once the existing running recalculation catches up.

peppy (approved):

peppy:
> it's just pulling data out of the existing model, not adding a new query on its own.

I meant for the fix operation on existing scores. `RecalculateTotalScores` operates on already imported `solo_scores` data, but as far as I can tell this requires pulling missing data from the old tables, which is a slightly more complex fixup operation.

But I think you're implying that `RecalculateTotalScores` alone is enough to fix this?

smoogipoo:
So, I think the misunderstanding is the title of this PR. It's only missing for the purpose of calculating total score. It is correctly populated/present in solo_scores, that's handled here:

https://github.com/ppy/osu-queue-score-statistics/blob/14c3be7b72c941872280586640ddf938a9995a3a/osu.Server.Queues.ScorePump/Queue/ImportHighScoresCommand.cs#L405

peppy:
Okay, so the main remaining consideration is that `RecalculateTotalScores` may need to be run a second time on all newer scores, in order to cover any blind spot from the first run (ie. newly imported scores with missing `MaxCombo`)?

smoogipoo:
Correct, and that run should be fairly quick since there should be few such scores.

PR #129, opened 2023-01-04
Allow reprocessing performance totals of users even when existing pp is zero

peppy:
Not sure why this is the case, but it's weird.

### DIFF
--- a/osu.Server.Queues.ScorePump/Performance/PerformanceCommand.cs
+++ b/osu.Server.Queues.ScorePump/Performance/PerformanceCommand.cs
@@ -65,8 +65,7 @@ namespace osu.Server.Queues.ScorePump.Performance
                 {
                     var userStats = await DatabaseHelper.GetUserStatsAsync(userId, RulesetId, db);
 
-                    // Only process users with an existing rank_score.
-                    if (userStats!.rank_score_exp == 0)
+                    if (userStats == null)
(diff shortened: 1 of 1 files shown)


frenzibyte:
Out of curiosity, I've blamed through the lines and it seems this might've lost some context (https://github.com/ppy/osu-queue-score-statistics/commit/391e087f503b0225a834a288c73c4bb0ccfbc4ce  -> https://github.com/ppy/osu-queue-score-statistics/commit/2f424e1b989a51bd3750eb026cf0231ba682b031 -> https://github.com/ppy/osu-queue-score-statistics/commit/2f424e1b989a51bd3750eb026cf0231ba682b031 -> this PR). Here's the earliest version of this check before the couple of refactor commits linked above:
https://github.com/ppy/osu-queue-score-statistics/blob/391e087f503b0225a834a288c73c4bb0ccfbc4ce/osu.Server.Queues.ScorePump/Performance/PerformanceProcessor.cs#L255-L257

Which indicates that the purpose of this check was to not process scores for inactive/restricted users, which lazer-only users so happen to fall under this condition as well.

peppy:
Thanks for checking. This pretty much confirms that it's okay to make this change, as the restricted case is being handled here:

smoogipoo (approved):

PR #130, opened 2023-01-12
Add option to use realtime difficulty calculation

smoogipoo:
In order to try to resolve https://github.com/ppy/osu/issues/21815, I want to try out performing realtime difficulty calculation to see where bottlenecks appear.

This PR adds two environment variables:
- [ ] `REALTIME_DIFFICULTY` : Set this to `0` to disable it and return to previous functionality (database lookups). Defaults to enabled if not set.
- [ ] Set envvar to `0` ^^.
- [ ] Set envvar to `0` ^^^^.
- `BEATMAP_DOWNLOAD_PATH` : Set this to the `osu` file download path. Defaults to `https://osu.ppy.sh/osu/{0}`.

Of particular importance, `BEATMAP_DOWNLOAD_PATH` should be set to a path that isn't rate-limited, similar to what is done for `osu-difficulty-calculator` and `osu-beatmap-difficulty-lookup-cache`.

**This PR contains no caching** - please advise on whether/how I should consider some form of minimal caching, perhaps at a `WorkingBeatmap` level, though I'm not sure how useful it'll be if the general case is players playing a very wide range of beatmaps.

**I have not considered full reprocesses** - we haven't done these in general yet so some caching and potentially reordering of operations will have to change in order to do that efficiently. This PR **does apply** to full reprocesses.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor.Tests/DatabaseTest.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor.Tests/DatabaseTest.cs
@@ -39,7 +39,7 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Tests
         protected DatabaseTest()
         {
             Environment.SetEnvironmentVariable("SCHEMA", "1");
-            Environment.SetEnvironmentVariable("INLINE_DIFFICULTY_CALCULATION", "0");
+            Environment.SetEnvironmentVariable("REALTIME_DIFFICULTY", "0");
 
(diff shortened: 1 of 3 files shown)


peppy:
@smoogipoo this is still fine to get in right?

peppy (approved):

smoogipoo:
Yeah it's still fine, but definitely remember to set that env variable. It's definitely not ready to be used for full recalcs and may not hold up for even queue processing.

peppy:
Yep, has been set.

PR #132, opened 2023-03-11
Fix build cache never being updated until server restart

peppy:
While investigating [reports of pp totals not updating](https://github.com/ppy/osu/discussions/22728#discussioncomment-5274115) I noticed that new builds may no be picked up for total calculations.

The fix is a very naive method of refreshing the build/blacklist cache.

I chose 1 minute to generally give it enough time to revalidate cache before someone was to complete a beatmap on a new release. We could potentially make this longer (or smarter) in the future if required.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserTotalPerformanceProcessor.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserTotalPerformanceProcessor.cs
@@ -23,6 +23,7 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Processors
         private BeatmapStore? beatmapStore;
         private BuildStore? buildStore;
 
+        // ReSharper disable once PrivateFieldCanBeConvertedToLocalVariable
         private readonly Timer cacheResetTimer;
 
         private static readonly bool process_user_totals = Environment.GetEnvironmentVariable("PROCESS_USER_TOTALS") != "0";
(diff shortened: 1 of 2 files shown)


peppy:
I don't know why but I could not see the failing inspection locally. Either something very weird going on, or jetbrains have changed something in an EAP release.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserTotalPerformanceProcessor.cs
```diff
@@ -22,11 +23,28 @@ public class UserTotalPerformanceProcessor : IProcessor
...
         public int Order => ScorePerformanceProcessor.ORDER + 1;
 
+        public UserTotalPerformanceProcessor()
+        {
+            cacheResetTimer = new Timer(60000);
+            cacheResetTimer.Elapsed += (_, _) =>
+            {
+                // naive cache clearing, mainly to ensure new builds are picked up on new client releases.
+                beatmapStore = null;
+                buildStore = null;
```
smoogipoo, line 41: I think this will cause nullrefs because it's not locked with the usages.

smoogipoo (changes requested):

osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserTotalPerformanceProcessor.cs
peppy: Is a very valid point.

peppy:
Went for a different approach which avoids the timer faffery as well.

smoogipoo (approved):

PR #135, opened 2023-06-28
Convert legacy total score to standardised when importing high scores

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu/pull/24072
- [x] Depends on `osu.Game` bump

One questionable/important aspect is that we don't store difficulty attributes for _all_ mod combinations. For example, for +HDHR, we only store the difficulty attributes for +HR. Because of this, there's an extra step here that re-multiplies the `LegacyComboScore` with the correct mod multiplier.  

This could result in an off-by-one score for each hitobject because [legacy scoring truncates at every judgement](https://github.com/ppy/osu/pull/24072/files#diff-c39a8a7ee2d48bcf35309f47a10b371f662af92a1b7638a0f7320d31591fa8edR162), that adds up over time. For example, [chocomint's score on `FREEDOM DiVE [FOUR DIMENSIONS]`](https://osu.ppy.sh/scores/osu/2177560145) converts to `1116538` when imported via this process, but `1116540` when imported directly from replay into lazer.

I've taken the liberty to treat this small discrepancy as acceptable.

---

Required database attributes:

```sql
INSERT INTO osu_difficulty_attribs (attrib_id, name) VALUES (23, 'Legacy Accuracy Score'), (25, 'Legacy Combo Score'), (27, 'Legacy Bonus Score Ratio');
```

We must re-run the difficulty calculator across all ranked beatmaps on all rulesets prior to any new import, with the osu-side changes applied.

### DIFF
--- a/osu.Server.Queues.ScorePump/RecalculateScoresCommand.cs
+++ b/osu.Server.Queues.ScorePump/RecalculateScoresCommand.cs
@@ -12,7 +12,6 @@ using McMaster.Extensions.CommandLineUtils;
 using osu.Game.Rulesets;
 using osu.Game.Rulesets.Judgements;
 using osu.Game.Rulesets.Scoring;
-using osu.Game.Scoring;
 using osu.Server.Queues.ScoreStatisticsProcessor;
 using osu.Server.Queues.ScoreStatisticsProcessor.Models;
 
@@ -58,7 +57,6 @@ namespace osu.Server.Queues.ScorePump
                         foreach (var score in scores)
                         {
                             bool requiresUpdate = ensureMaximumStatistics(score);
-                            requiresUpdate |= ensureCorrectTotalScore(score);
 
                             if (requiresUpdate)
                             {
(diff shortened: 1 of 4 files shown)


bdach:
Just for reference, as far as I can tell, this will also fix an issue reported on discord today about [realtime imports of stable scores not having mod multipliers applied](https://discord.com/channels/188630481301012481/1097318920991559880/1131261929386672158). Current `master` does not include mods when computing total score from reference, while this pull will do, because game-side `StandardisedScoreMigrationTools` do.

smoogipoo:
@peppy I want to test SR/PP changes https://github.com/ppy/osu/pull/22613 and https://github.com/ppy/osu/pull/20963 and I need this PR + https://github.com/ppy/osu-difficulty-calculator/pull/221 merged otherwise it gets messy dealing with multiple working trees.

osu.Server.Queues.ScorePump/RecalculateScoresCommand.cs
```diff
@@ -58,7 +57,6 @@ public async Task<int> OnExecuteAsync(CancellationToken cancellationToken)
                         foreach (var score in scores)
                         {
                             bool requiresUpdate = ensureMaximumStatistics(score);
-                            requiresUpdate |= ensureCorrectTotalScore(score);
```
peppy, line 61: Does this command even make sense anymore?

peppy:
Haven't tested running this, but I think we can move forward for now.

peppy (approved):

PR #139, opened 2023-08-16
Combine projects, tidy up namespaces, update README

peppy:
Previously, this project was split in two pieces. It made more sense years ago, and zero sense now.

- To run the queue processor in processing mode, you now use `dotnet run -- queue run`
- There's no more "score pump" project
- I've also removed some commands which were single use / obsoleted.

Would recommend reading the updated `README` for a better summary of this project.

In the future I'd probably hope to rename this project to `osu-score-statistics-processor` since it's not always a queue processor, but don't want to stir too much for now.


smoogipoo (approved):

ilsubyeega:
By the way this pr changes `osu.Server.Queues.ScorePump/Queue/ImportHighScoresCommand.cs` to `osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/ImportHighScoresCommand.cs`, the document https://github.com/ppy/osu-infrastructure/blob/master/score-submission.md?plain=1#L337 should be relinked to the latest version.

peppy:
Thanks, I've updated the link.

Issue #141, opened 2023-08-21
Add method for cleaning up non-high scores


peppy:
ie. scores with `preserve = 0` set (and meeting various other conditions).

peppy:
As a note here, depending on https://github.com/ppy/osu-infrastructure/issues/16's conclusion this command will need to be altered.

Put another way, we need to make sure that partition rotation doesn't happen to soon, to ensure that things like replay cleanup can be performed first.

peppy:
Short term, this won't be running. We'll keep all scores around until it's ready to go.

Givikap120:
Considering that No Combo Scaling rework is on the way - it should be considered that scores, that can potentially be new pp top play on map, will be saved.
For example - 1 miss in the middle of the map can have less score with choke with 4 misses in the end, but in new pp system it will have more pp.
So that means that saved scores should be saved in the way that most of the situations like this will be handled correctly (i understand that covering all is kinda impossible).
And of course mod-unique score saving should be preserved (because it's by far the most common situation with less score more pp).

bdach:
@peppy do you possibly recall what the last task item here ("Add queue to signal that a score may not need to be preserved anymore") means? Who's this queue for?

peppy:
I believe that was for osu-web to push to (ie. when a score was unpinned) so we could handle accordingly.

Right now, how we run clean-up is a bit up-in-the-air. My current plan is to run on all users which have recent scores, iterating all their scores. Which may work okay but also may take too much processing power, not sure yet. Could still use a queue for this though, and just push all recent user ids to the queue.

Not sure yet.

bdach:
I see, thanks. I guess I'll leave that last item alone for now then?

There's one further web change I wanna make related to this issue that wasn't noted down before, just a little something I noticed in passing. Intent is to get it out today.

peppy:
This is fully completed with merge of https://github.com/ppy/osu-queue-score-statistics/pull/392 🥳

PR #144, opened 2023-08-24
Add command to clean up non-preserved scores

peppy:
Note that cancellation doesn't work because dapper. Will probably want to consider migrating to something that does work.

I'm not aiming for efficiency here because it doesn't need to be that efficient. Currently running on production to test (is cleaning around 500 scores per second).

Addresses part of https://github.com/ppy/osu-queue-score-statistics/issues/141 (deleting of scores which are already in a non-preserved state). Remaining piece is to downgrade preserved scores to non-preserved when they can be.


smoogipoo (approved):

peppy:
As expected, streaming the results fails with usual shit. Not sure if ProxySQL or something else:

`Unhandled exception. System.IO.EndOfStreamException: Expected to read 692 payload bytes but only received 477.`

PR #148, opened 2023-08-30
Add ability to re-run score imports and update existing items

peppy:
RFC at best. Will need to backport this to an older release and test before it's ready for review.


peppy:
Have tested this to work.

peppy:
For reference, I've pushed https://github.com/peppy/osu-queue-score-statistics/tree/production-importer which has these changes rebased on the correct version for (current) production processing.

I also pushed a package to docker (still scorepump) with these changes: https://hub.docker.com/layers/pppy/osu-queue-score-pump/production-import/images/sha256-4735ae645fbb25360c5b1d0671178fb23a07f74ebef02f26883c5ba208c52868?context=explore

PR #149, opened 2023-08-30
Add support for increasing rank counts

### LINKED ISSUE #133: Add support for increasing rank counts (S / SS / A / etc)

bdach:
Closes #133.

As per [discord conversation](https://discord.com/channels/188630481301012481/188630652340404224/1146035221628731423), basically written with zero regard for performance, only correctness of behaviour. There are probably places that are going to hurt in this implementation.

Notably, this implementation relies on there being an absolute ordering of scores from a given user on a given beatmap in a given ruleset, to determine which score to pick as the "source" of rank to tally. The criteria chosen are: total score descending, then ID descending (as a substitute for "newest") as a tie-breaker.

[I brought that up on discord too](https://discord.com/channels/188630481301012481/188630652340404224/1146055765107421254), and there were two alternatives offered for the tie-breaker: either highest rank, or newest score, whichever easiest. I wanted to do highest rank but for practical reasons this is more difficult than you'd think (tl;dr: can't query rank directly from SQL, have to reach into the `` `solo_scores`.`data` `` JSON to get it, wherein it's a string and not the enum, so I'd have to sort it...), therefore the choice of ID instead.

No idea if we want to be giving any considerations to mods here.

Do be careful with review on this one, it's the first time I'm touching this project to do anything serious.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserRankCountProcessor.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserRankCountProcessor.cs
@@ -1,6 +1,7 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using System.Diagnostics;
 using Dapper;
 using JetBrains.Annotations;
 using MySqlConnector;
@@ -18,24 +19,26 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Processors
     {
         public void RevertFromUserStats(SoloScoreInfo score, UserStats userStats, int previousVersion, MySqlConnection conn, MySqlTransaction transaction)
         {
+            if (!score.Passed)
+                return;
+
             if (previousVersion >= 7)
             {
-                // note that this implementation is only supposed to be correct if a revert is immediately succeeded by a reprocessing of the same score.
-                // at the time of writing this is the only case where this method is called, but if that ever changes this will likely need reconsideration.
+                // It is assumed that in the case of a revert, either the score is deleted, or a reapplication will immediately follow.
 
-                // first, see if the score we're reverting is the user's best (and as such included in the rank counts).
-                var bestScore = getBestScore(score.UserID, score.BeatmapID, score.RulesetID, excludedScoreId: null, conn, transaction);
+                // First, see if the score we're reverting is the user's best (and as such included in the rank counts).
+                var bestScore = getBestScore(score, conn, transaction);
 
                 if (bestScore?.ID != score.ID)
                 {
-                    // this score isn't the user's best on the beatmap, so nothing needs to be reverted.
+                    // This score isn't the user's best on the beatmap, so nothing needs to be reverted.
                     return;
                 }
 
                 updateRankCounts(userStats, score.Rank, revert: true);
 
-                // this score is the user's best, so fetch the next best (excluding it) so that we can apply the rank from that score.
-                var secondBestScore = getBestScore(score.UserID, score.BeatmapID, score.RulesetID, excludedScoreId: score.ID, conn, transaction);
+                // This score is the user's best, so fetch the next best so that we can apply the rank from that score.
+                var secondBestScore = getSecondBestScore(score, conn, transaction);
                 if (secondBestScore != null)
                     updateRankCounts(userStats, secondBestScore.Rank, revert: false);
             }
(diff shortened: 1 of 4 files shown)


osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserRankCountProcessor.cs
```diff
@@ -0,0 +1,118 @@
...
+            // purpose of join with `process_history` is to only look for already-processed scores
+            var rankSource = conn.QueryFirstOrDefault<SoloScore?>(
+                $"SELECT * FROM {SoloScore.TABLE_NAME} `s`"
+                + $"JOIN {ProcessHistory.TABLE_NAME} `ph` ON `ph`.`score_id` = `s`.`id`"
+                + "WHERE (@ExcludedScoreId IS NULL OR `s`.`id` <> @ExcludedScoreId) "
+                + "AND `s`.`user_id` = @UserId "
+                + "AND `s`.`beatmap_id` = @BeatmapId "
+                + "AND `s`.`ruleset_id` = @RulesetId "
+                + "AND `s`.`preserve` = 1 "
+                + "ORDER BY JSON_EXTRACT(`s`.`data`, '$.total_score') DESC, `s`.`id` DESC", new
```
bdach, line 79: This `JSON_EXTRACT()` is probably not going to be a good time. Not sure what should happen (extract to column? some other mysql magic? on MSSQL I'd slap on a computed column and an index on that, but that's not a thing on mysql apparently)

peppy: It's fine, since we're already filtering down to per-beatmap scores. Generally we should be working with well under 100 scores at this point.

peppy, line 79:
```suggestion
                + "ORDER BY `s`.`data`->'$.total_score' DESC, `s`.`id` DESC LIMIT 1", new
```

I think the shorthand json extract syntax is nicer. Also I'm not 100% (should probably test) sure, but I think the query will return less data and therefore be marginally faster if `LIMIT` is applied.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserRankCountProcessor.cs
```diff
@@ -0,0 +1,118 @@
...
+        {
+        }
+
+        private static SoloScoreInfo? getBestScore(int userId, int beatmapId, int rulesetId, ulong? excludedScoreId, MySqlConnection conn, MySqlTransaction transaction)
+        {
+            // purpose of join with `process_history` is to only look for already-processed scores
+            var rankSource = conn.QueryFirstOrDefault<SoloScore?>(
+                $"SELECT * FROM {SoloScore.TABLE_NAME} `s`"
+                + $"JOIN {ProcessHistory.TABLE_NAME} `ph` ON `ph`.`score_id` = `s`.`id`"
+                + "WHERE (@ExcludedScoreId IS NULL OR `s`.`id` <> @ExcludedScoreId) "
```
peppy, line 74:
```suggestion
                + "WHERE (@ExcludedScoreId IS NULL OR `s`.`id` != @ExcludedScoreId) "
```

I generally prefer `!=` to match what we're used to.

peppy (commented):
I think this looks mostly correct, but will need to figure index concerns before considering merging.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserRankCountProcessor.cs
bdach: > I think the shorthand json extract syntax is nicer.

That was a "didn't read docs far enough" situation. And also style I guess. Applied, though :+1: 

> Also I'm not 100% (should probably test) sure, but I think the query will return less data and therefore be marginally faster if `LIMIT` is applied.

Looking in `general_log` dapper seems to set `SET sql_select_limit=1;` before the query and restore it to `NULL` after, but I've applied for safety :+1:

peppy:
I've made some changes here, as discussed on stream:

- Fixed non-pass scores not being considered correctly (failing test coverage on previous version in https://github.com/ppy/osu-queue-score-statistics/pull/149/commits/50a10c843d9c9d1043bfe9d97093811d0d2958a9)
- Refactored the process to avoid the `excludedScoreId` as I found it a bit hard to mentally parse.
- Have tested that the updated query does run efficiently against the final table / index / partition structure.

@bdach Please double check you're on board with my changes. If so, this can be merged. We are still running an older version of this processor on production due to diffcalc considerations, so there's no reason not to get this in ready for when we're ready to deploy at scale.

peppy (approved):

osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserRankCountProcessor.cs
```diff
@@ -18,76 +19,84 @@ public class UserRankCountProcessor : IProcessor
...
-                + "AND `s`.`ruleset_id` = @RulesetId "
-                + "AND `s`.`preserve` = 1 "
-                + "ORDER BY `s`.`data`->'$.total_score' DESC, `s`.`id` DESC "
-                + "LIMIT 1", new
+                "SELECT * FROM solo_scores WHERE `user_id` = @user_id "
+                + "AND `beatmap_id` = @beatmap_id "
+                + "AND `ruleset_id` = @ruleset_id "
+                // preserve is not flagged on the newly arriving score until it has been completely processed (see logic in `ScoreStatisticsQueueProcessor.cs`)
+                // therefore we need to make an exception here to ensure it's included.
+                + "AND `preserve` = 1 OR `id` = @new_score_id "
```
bdach, line 80: I'm... not actually sure on the correctness of precedence here? The intent seems to be

```suggestion
                + "AND (`preserve` = 1 OR `id` = @new_score_id) "
```

and I'd probably feel much safer with the extra set of parens here even if that's how it ends up being executed...

cl8n: AND comes before OR, so the parenthesis do make a difference.

peppy: I'm not sure where they went. I had them there when testing the query for correctness.

peppy: Right, I removed them to test whether they were causing this rider fuckup:

JetBrains Rider 2023-09-13 at 09 39 52

Just to confirm, all testing was done *with* the parens.

developomp:
:tada:

PR #150, opened 2023-08-30
Fix mod multipliers during legacy score conversion

smoogipoo:
Prereqs:
- [ ] https://github.com/ppy/osu/pull/24685

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/ImportHighScoresCommand.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/ImportHighScoresCommand.cs
@@ -545,8 +545,19 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Queue
                 // To handle this case, the combo-multiplied portion is readjusted with the new mod multiplier.
                 if (difficultyMods != highScore.enabled_mods)
                 {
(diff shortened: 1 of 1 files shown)


peppy:
@bdach are you okay with this one? Unsure if you performed a review here yet.

bdach:
Haven't looked through this yet no.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/ImportHighScoresCommand.cs
```diff
@@ -575,14 +576,21 @@ private async Task<ScoreInfo> createReferenceScore(Ruleset ruleset, HighScore hi
                     scoreInfo.MaximumStatistics[HitResult.LegacyComboIncrease] = (int)maxComboAttribute.value - maxComboFromStatistics;
 #pragma warning restore CS0618
 
+                Beatmap beatmap = await connection.QuerySingleAsync<Beatmap>("SELECT * FROM osu_beatmaps WHERE `beatmap_id` = @BeatmapId", new
+                {
+                    BeatmapId = highScore.beatmap_id
+                }, transaction);
+
+                LegacyBeatmapConversionDifficultyInfo difficulty = LegacyBeatmapConversionDifficultyInfo.FromAPIBeatmap(beatmap.ToAPIBeatmap());
```
bdach, line 584: Ascertaining that this construction works completely correctly (especially with respect to ruleset) is a really roundabout trip, but it does appear that this should do what is expected...

bdach (approved):

bdach:
Seems okay source-wise, but warning that I have not tested this as I'm not even sure how one _would_ test this.

PR #151, opened 2023-08-30
Replace all usage of `TABLE_NAME` constants with direct table names

peppy:
Rider can't understand these constants, which makes database introspection quite useless. Also I don't think it's making things any clearer having the constants.


smoogipoo:
I also find this easier to read these days :p Just a few inspections to fix

smoogipoo (approved):

PR #153, opened 2023-09-05
Fix new highscores not being imported

smoogipoo:
The fix here is to use an object that is a `class`, which can be nullable unlike a non-nullable tuple type. The latter is uglier to make work.

Context:
- I run this to import highscores to solo scores: https://github.com/smoogipoo/diffcalc-sheet-generator/blob/f0319ea34e27791091ab212ca690e9fa1c3db224/docker/importer/start.sh#L29-L38
- `SkipExisting` is true by default.
- Because it's an array of `(ulong, ulong)` which is non-nullable, the first condition `if ((existingMapping != null && skipExisting) ...` passes and the score is skipped.


smoogipoo:
I don't know if this affects anything in prod/dev, but you may need to backport @peppy .

smoogipoo:
Also @peppy I don't know where the backported changes even are... I heard above that you backported https://github.com/ppy/osu-queue-score-statistics/pull/148 but I can't find it on any branch.

peppy (approved):

PR #154, opened 2023-09-05
Add command to migrate playlist scores from `multiplayer_scores` to `solo_scores`

peppy:
This is to be used to preserve ongoing featured artist / spotlight room's scores as we migrate scores to the new table.

It means we can continue to move forward without worrying about breaking things, basically.

It's made to only run for playlists, and not migrate *all* scores, but could be expanded in the future if we want to have more coverage.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/MigratePlaylistScoresToSoloScoresCommand.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/MigratePlaylistScoresToSoloScoresCommand.cs
@@ -1,0 +1,190 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+using System.Diagnostics;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Dapper;
+using Dapper.Contrib.Extensions;
+using McMaster.Extensions.CommandLineUtils;
+using Newtonsoft.Json;
+using osu.Game.Online.API;
+using osu.Game.Online.API.Requests.Responses;
+using osu.Game.Rulesets;
+using osu.Game.Rulesets.Scoring;
+using osu.Game.Scoring;
+using osu.Server.Queues.ScoreStatisticsProcessor.Helpers;
+using osu.Server.Queues.ScoreStatisticsProcessor.Models;
+
+namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Maintenance
+{
+    [Command("migrate-playlist-scores", Description = "Migrate scores from `multiplayer_scores` to `solo_scores`.")]
+    public class MigratePlaylistScoresToSoloScoresCommand : BaseCommand
+    {
+        /// <summary>
+        /// The playlist room ID to reprocess.
+        /// </summary>
+        [Required]
+        [Argument(0, Description = "Command separated list of playlist room IDs to reprocess.")]
+        public string PlaylistIds { get; set; } = string.Empty;
+
+        public async Task<int> OnExecuteAsync(CancellationToken cancellationToken)
+        {
+            foreach (string id in PlaylistIds.Split(','))
+            {
+                if (cancellationToken.IsCancellationRequested)
+                    break;
+
+                using (var db = Queue.GetDatabaseConnection())
+                {
+                    var playlistItems = await db.QueryAsync<MultiplayerPlaylistItem>("SELECT * FROM multiplayer_playlist_items WHERE room_id = @PlaylistId", new
+                    {
+                        PlaylistId = int.Parse(id),
+                    });
+
+                    foreach (var item in playlistItems)
+                    {
+                        if (cancellationToken.IsCancellationRequested)
+                            break;
+
+                        MultiplayerScore[] scores = (await db.QueryAsync<MultiplayerScore>($"SELECT * FROM multiplayer_scores WHERE playlist_item_id = {item.id}")).ToArray();
+
+                        foreach (var score in scores)
+                        {
+                            Console.WriteLine($"Reprocessing score {score.id} from playlist {item.id}");
+
+                            Ruleset ruleset = LegacyRulesetHelper.GetRulesetFromLegacyId(item.ruleset_id);
+                            HitResult maxRulesetJudgement = ruleset.GetHitResults().First().result;
+                            Dictionary<HitResult, int> statistics = JsonConvert.DeserializeObject<Dictionary<HitResult, int>>(score.statistics)!;
+                            List<HitResult> allHits = statistics
+                                                      .SelectMany(kvp => Enumerable.Repeat(kvp.Key, kvp.Value))
+                                                      .ToList();
+                            var maximumStatistics = new Dictionary<HitResult, int>();
+
+                            foreach (var groupedStats in allHits
+                                                         .Select(r => getMaxJudgementFor(r, maxRulesetJudgement))
+                                                         .GroupBy(r => r))
+                            {
+                                maximumStatistics[groupedStats.Key] = groupedStats.Count();
+                            }
+
+                            APIMod[] roomMods = JsonConvert.DeserializeObject<APIMod[]>(item.required_mods)!;
+                            APIMod[] scoreMods = JsonConvert.DeserializeObject<APIMod[]>(score.mods)!;
+
+                            foreach (var m in roomMods)
+                                Debug.Assert(scoreMods.Contains(m));
+
+                            Console.WriteLine($"Mods: {string.Join(',', scoreMods.Select(m => m.ToString()))}");
+
+                            int insertId = await db.InsertAsync(new SoloScore
+                            {
+                                user_id = (int)score.userId,
+                                beatmap_id = (int)score.beatmapId,
+                                ruleset_id = item.ruleset_id,
+                                preserve = true,
+                                ScoreInfo = SoloScoreInfo.ForSubmission(new ScoreInfo
+                                {
+                                    Statistics = statistics,
+                                    MaximumStatistics = maximumStatistics,
+                                    Ruleset = ruleset.RulesetInfo,
+                                    MaxCombo = score.max_combo,
+                                    APIMods = scoreMods,
+                                    Accuracy = (double)score.accuracy!,
+                                    TotalScore = (long)score.total_score!,
+                                    Rank = Enum.Parse<ScoreRank>(score.rank),
+                                    Passed = score.passed,
+                                }),
+                                created_at = (DateTimeOffset)score.created_at!,
+                                updated_at = (DateTimeOffset)score.updated_at!,
+                            });
+
+                            await db.ExecuteAsync("INSERT INTO multiplayer_score_links (user_id, room_id, beatmap_id, playlist_item_id, score_id, created_at, updated_at) VALUES (@userId, @roomId, @beatmapId, @playlistItemId, @scoreId, @createdAt, @updatedAt)", new
+                            {
+                                score.userId,
+                                score.roomId,
+                                score.beatmapId,
+                                score.playlistItemId,
+                                scoreId = insertId,
+                                createdAt = score.started_at,
+                                updatedAt = score.ended_at
+                            });
+
+                            Console.WriteLine();
+                        }
+                    }
+                }
+            }
+
+            Console.WriteLine("Finished.");
+            return 0;
+        }
+
+        private static HitResult getMaxJudgementFor(HitResult hitResult, HitResult max)
+        {
+            switch (hitResult)
+            {
+                case HitResult.Miss:
+                case HitResult.Meh:
+                case HitResult.Ok:
+                case HitResult.Good:
+                case HitResult.Great:
+                case HitResult.Perfect:
+                    return max;
+
+                case HitResult.SmallTickMiss:
+                case HitResult.SmallTickHit:
+                    return HitResult.SmallTickHit;
+
+                case HitResult.LargeTickMiss:
+                case HitResult.LargeTickHit:
+                    return HitResult.LargeTickHit;
+            }
+
+            return HitResult.IgnoreHit;
+        }
+
+        // ReSharper disable InconsistentNaming
+        public class MultiplayerScore
+        {
+            public long id { get; set; }
+            public uint userId { get; set; }
+            public long roomId { get; set; }
+            public long playlistItemId { get; set; }
+            public long beatmapId { get; set; }
+            public string rank { get; set; } = string.Empty;
+            public long? total_score { get; set; }
+            public double? accuracy { get; set; }
+            public double? pp { get; set; }
+            public int max_combo { get; set; }
+            public string mods { get; set; } = string.Empty;
+            public string statistics { get; set; } = string.Empty;
+            public DateTime started_at { get; set; }
+            public DateTime? ended_at { get; set; }
+            public bool passed { get; set; }
+            public DateTime? created_at { get; set; }
+            public DateTime? updated_at { get; set; }
+            public DateTime? deleted_at { get; set; }
+        }
+
+        public class MultiplayerPlaylistItem
+        {
+            public long id { get; set; }
+            public long room_id { get; set; }
+            public uint owner_id { get; set; }
+            public uint beatmap_id { get; set; }
+            public ushort ruleset_id { get; set; }
+            public ushort? playlist_order { get; set; }
+            public string allowed_mods { get; set; } = string.Empty;
+            public string required_mods { get; set; } = string.Empty;
+            public byte? max_attempts { get; set; }
+            public DateTime? created_at { get; set; }
+            public DateTime? updated_at { get; set; }
+            public bool expired { get; set; }
+            public DateTime? played_at { get; set; }
+        }
+    }
+}
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/MigratePlaylistScoresToSoloScoresCommand.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/MigratePlaylistScoresToSoloScoresCommand.cs
@@ -80,7 +80,7 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Maintenance
 
                             int? insertId = null;
 
-                            if (score.total_score != null)
+                            if (score.total_score != null || score.ended_at != null)
                             {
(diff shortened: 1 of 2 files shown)


osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/MigratePlaylistScoresToSoloScoresCommand.cs
```diff
@@ -0,0 +1,190 @@
...
+            public bool passed { get; set; }
+            public DateTime? created_at { get; set; }
+            public DateTime? updated_at { get; set; }
+            public DateTime? deleted_at { get; set; }
+        }
+
+        public class MultiplayerPlaylistItem
+        {
+            public long id { get; set; }
+            public long room_id { get; set; }
```
bdach, line 176: Shouldn't these be

```suggestion
            public ulong id { get; set; }
            public ulong room_id { get; set; }
```

At least that's what I get from

```
MySQL [osu]> describe multiplayer_playlist_items;
+----------------+--------------------+------+-----+---------+----------------+
| Field          | Type               | Null | Key | Default | Extra          |
+----------------+--------------------+------+-----+---------+----------------+
| id             | bigint unsigned    | NO   | PRI | NULL    | auto_increment |
| room_id        | bigint unsigned    | NO   | MUL | NULL    |                |
| owner_id       | int unsigned       | NO   |     | NULL    |                |
| beatmap_id     | mediumint unsigned | NO   |     | NULL    |                |
| ruleset_id     | smallint unsigned  | YES  |     | NULL    |                |
| playlist_order | smallint unsigned  | YES  |     | NULL    |                |
| allowed_mods   | json               | YES  |     | NULL    |                |
| required_mods  | json               | YES  |     | NULL    |                |
| max_attempts   | tinyint unsigned   | YES  |     | NULL    |                |
| created_at     | timestamp          | YES  |     | NULL    |                |
| updated_at     | timestamp          | YES  |     | NULL    |                |
| expired        | tinyint(1)         | NO   |     | 0       |                |
| played_at      | timestamp          | YES  |     | NULL    |                |
+----------------+--------------------+------+-----+---------+----------------+
13 rows in set (0.002 sec)
```

In practice it shouldn't matter for a _very_ long time, but the remaining columns have signedness set correctly, and I'd rather be a stickler for these given how sensitive dapper is to these...

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/MigratePlaylistScoresToSoloScoresCommand.cs
```diff
@@ -0,0 +1,190 @@
...
+            public uint owner_id { get; set; }
+            public uint beatmap_id { get; set; }
+            public ushort ruleset_id { get; set; }
```
bdach, line 179: Hmm in my DBs this is nullable / `ushort?`... Is production different? If not, may be worth running a

```sql
SELECT COUNT(1) FROM multiplayer_playlist_items WHERE ruleset_id IS NULL;
```

to check for any surprises there, because otherwise scores may accidentally get misattributed to osu!-the-ruleset... I don't have any rows with `NULL` in my dbs, but that means precisely nothing.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/MigratePlaylistScoresToSoloScoresCommand.cs
```diff
@@ -0,0 +1,190 @@
...
+        }
+
+        // ReSharper disable InconsistentNaming
+        public class MultiplayerScore
+        {
+            public long id { get; set; }
+            public uint userId { get; set; }
+            public long roomId { get; set; }
+            public long playlistItemId { get; set; }
+            public long beatmapId { get; set; }
```
bdach, line 157: Similarly:

```suggestion
            public ulong id { get; set; }
            public uint userId { get; set; }
            public ulong roomId { get; set; }
            public ulong playlistItemId { get; set; }
            public uint beatmapId { get; set; }
```

from

```
MySQL [osu]> describe multiplayer_scores;
+------------------+---------------------------------------------+------+-----+-------------------+-------------------+
| Field            | Type                                        | Null | Key | Default           | Extra             |
+------------------+---------------------------------------------+------+-----+-------------------+-------------------+
| id               | bigint unsigned                             | NO   | PRI | NULL              | auto_increment    |
| user_id          | int unsigned                                | NO   | MUL | 0                 |                   |
| room_id          | bigint unsigned                             | NO   | MUL | NULL              |                   |
| playlist_item_id | bigint unsigned                             | NO   | MUL | NULL              |                   |
| beatmap_id       | mediumint unsigned                          | NO   |     | NULL              |                   |
| rank             | enum('A','B','C','D','S','SH','X','XH','F') | YES  |     | NULL              |                   |
| total_score      | mediumint unsigned                          | YES  |     | NULL              |                   |
| accuracy         | double(5,4)                                 | YES  |     | NULL              |                   |
| pp               | double(8,2)                                 | YES  |     | NULL              |                   |
| max_combo        | mediumint unsigned                          | YES  |     | NULL              |                   |
| mods             | json                                        | YES  |     | NULL              |                   |
| statistics       | json                                        | YES  |     | NULL              |                   |
| started_at       | timestamp                                   | NO   |     | CURRENT_TIMESTAMP | DEFAULT_GENERATED |
| ended_at         | timestamp                                   | YES  |     | NULL              |                   |
| passed           | tinyint(1)                                  | YES  |     | NULL              |                   |
| created_at       | timestamp                                   | YES  |     | NULL              |                   |
| updated_at       | timestamp                                   | YES  |     | NULL              |                   |
| deleted_at       | timestamp                                   | YES  |     | NULL              |                   |
+------------------+---------------------------------------------+------+-----+-------------------+-------------------+
18 rows in set (0.002 sec)
```

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/MigratePlaylistScoresToSoloScoresCommand.cs
```diff
@@ -0,0 +1,190 @@
...
+            public string rank { get; set; } = string.Empty;
+            public long? total_score { get; set; }
+            public double? accuracy { get; set; }
+            public double? pp { get; set; }
+            public int max_combo { get; set; }
```
bdach, line 162:
```suggestion
            public uint? max_combo { get; set; }
```

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/MigratePlaylistScoresToSoloScoresCommand.cs
```diff
@@ -0,0 +1,190 @@
...
+                                                      .SelectMany(kvp => Enumerable.Repeat(kvp.Key, kvp.Value))
+                                                      .ToList();
+                            var maximumStatistics = new Dictionary<HitResult, int>();
+
+                            foreach (var groupedStats in allHits
+                                                         .Select(r => getMaxJudgementFor(r, maxRulesetJudgement))
+                                                         .GroupBy(r => r))
+                            {
+                                maximumStatistics[groupedStats.Key] = groupedStats.Count();
+                            }
```
bdach, line 73: Not sure how much speed matters here, but I do find this implementation and the way it materialises `allHits` a bit wasteful, when instead you could do

```csharp
foreach (var resultCount in statistics)
{
    var maxResult = getMaxJudgementFor(resultCount.Key, maxRulesetJudgement);

    if (!statistics.ContainsKey(maxResult))
        statistics[maxResult] = resultCount.Value;
    else
        statistics[maxResult] += resultCount.Value;
}
```

Both _work_ though, so if this works well enough for you, this is a dismissable concern.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/MigratePlaylistScoresToSoloScoresCommand.cs
```diff
@@ -0,0 +1,190 @@
...
+
+                            APIMod[] roomMods = JsonConvert.DeserializeObject<APIMod[]>(item.required_mods)!;
+                            APIMod[] scoreMods = JsonConvert.DeserializeObject<APIMod[]>(score.mods)!;
+
+                            foreach (var m in roomMods)
+                                Debug.Assert(scoreMods.Contains(m));
+
+                            Console.WriteLine($"Mods: {string.Join(',', scoreMods.Select(m => m.ToString()))}");
```
bdach, line 81: Was this intended to be in the command output? Feels kinda random.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/MigratePlaylistScoresToSoloScoresCommand.cs
```diff
@@ -0,0 +1,190 @@
...
+                                ScoreInfo = SoloScoreInfo.ForSubmission(new ScoreInfo
+                                {
+                                    Statistics = statistics,
+                                    MaximumStatistics = maximumStatistics,
+                                    Ruleset = ruleset.RulesetInfo,
+                                    MaxCombo = score.max_combo,
+                                    APIMods = scoreMods,
+                                    Accuracy = (double)score.accuracy!,
+                                    TotalScore = (long)score.total_score!,
+                                    Rank = Enum.Parse<ScoreRank>(score.rank),
```
bdach, line 98: All of these are nullable in `multiplayer_scores`, so any of these could die in practice. Could probably happen if a client dies before submission or something. I definitely have such rows on my test db, although I cannot recall specific circumstances. Another thing to check for before attempting to run I guess?

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/MigratePlaylistScoresToSoloScoresCommand.cs
```diff
@@ -0,0 +1,190 @@
...
+
+                            Console.WriteLine($"Mods: {string.Join(',', scoreMods.Select(m => m.ToString()))}");
+
+                            int insertId = await db.InsertAsync(new SoloScore
+                            {
+                                user_id = (int)score.userId,
+                                beatmap_id = (int)score.beatmapId,
+                                ruleset_id = item.ruleset_id,
+                                preserve = true,
+                                ScoreInfo = SoloScoreInfo.ForSubmission(new ScoreInfo
```
bdach, line 89: Since `ForSubmission()` is [mostly intended as a client-side method](https://github.com/ppy/osu/blob/611933a029236fce3ec309f2f9da319f8a0405aa/osu.Game/Online/API/Requests/Responses/SoloScoreInfo.cs#L213-L225), there's some stuff it isn't populating but which could be backpopulated in a server-side context such as this. Looking at other `solo_scores` rows in my DB, fields that could be present in there that I don't believe will be with this implementation are:

- `ended_at`
- `beatmap_id` (arguably redundant)
- `started_at`

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/MigratePlaylistScoresToSoloScoresCommand.cs
nanaya: right, I forgot about the non-finished score case. Those should probably still have score_link entry for attempts count but no solo_score entry.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/MigratePlaylistScoresToSoloScoresCommand.cs
peppy: I don't think it should be nullable.

```sql

root@db-master:osu> SELECT COUNT(*) FROM multiplayer_playlist_items WHERE ruleset_id IS NULL;
+----------+
| COUNT(*) |
+----------+
| 0        |
+----------+
```

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/MigratePlaylistScoresToSoloScoresCommand.cs
peppy: This was left over from the command I copied from because I was unsure mods were being reserialised correctly. Can be removed.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/MigratePlaylistScoresToSoloScoresCommand.cs
peppy: I think what I have reads slightly better so I'll keep it for now.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/MigratePlaylistScoresToSoloScoresCommand.cs
peppy: We probably need to revisit what is being store in the json and what is necessary to be stored in there. I think the original thinking was that the JSON content should stand on its own to represent the play, but I dunno.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/MigratePlaylistScoresToSoloScoresCommand.cs
bdach: Okay great, just checking. Let's just say I'm being overly thorough in checking due to having an allergy for data anomalies from *prior bad experiences*...

nanaya: good catch, I believe this column can be just made `not null`. At least on osu-web side the field is required.

bdach: (just gonna crosslink https://github.com/ppy/osu-web/pull/10523 for reference)

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/MigratePlaylistScoresToSoloScoresCommand.cs
```diff
@@ -0,0 +1,199 @@
...
+
+                                insertId = await db.InsertAsync(new SoloScore
+                                {
+                                    user_id = (int)score.userId,
+                                    beatmap_id = (int)score.beatmapId,
+                                    ruleset_id = item.ruleset_id,
+                                    preserve = true,
+                                    ScoreInfo = soloScoreInfo,
+                                    created_at = (DateTimeOffset)score.created_at!,
+                                    updated_at = (DateTimeOffset)score.updated_at!,
```
bdach, line 110: So this is another potential minefield... If the machine that this code is going to be running on will be running on the UTC timezone, this is gonna _work_. But if it's not, then this is gonna silently break.

As far as I can tell, `multiplayer_scores` is keeping dates in `created_at` and `updated_at` as UTC. _However_, because computers are cursed, `DateTime` instances returned by dapper have `DateTimeKind.Unspecified` set. Now the `DateTimeOffset` conversion [assumes that `Unspecified` dates are local](https://github.com/dotnet/runtime/blob/1487d46e4ff7731b52af92b1228373a2a66ed6bf/src/libraries/System.Private.CoreLib/src/System/DateTimeOffset.cs#L85-L89), mostly because it _has_ to flip a coin in such a case. But what that means, is that _if_ the machine that this is ran on doesn't use UTC timezone, then the datetimes stored to db will be _skewed back_ by the inverse of the local computer's timezone offset.

I _believe_ that this is a concern, then the more correct way to convert would be

```suggestion
                                    created_at = new DateTime(score.created_at!.Value.Ticks, DateTimeKind.Utc),
                                    updated_at = new DateTime(score.updated_at!.Value.Ticks, DateTimeKind.Utc),
```

But this _may_ be irrelevant and it's probably best that you double-check this, because this crap is _deeply_ confusing and screwed.

The same applies to `soloScoreInfo.{Started,Ended}At` above.

bdach (changes requested):
Notes from local testing

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/MigratePlaylistScoresToSoloScoresCommand.cs
peppy: Has been fixed.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/MigratePlaylistScoresToSoloScoresCommand.cs
```diff
@@ -0,0 +1,199 @@
...
+        }
+
+        // ReSharper disable InconsistentNaming
+        public class MultiplayerScore
+        {
+            public ulong id { get; set; }
+            public uint userId { get; set; }
+            public ulong roomId { get; set; }
+            public ulong playlistItemId { get; set; }
+            public uint beatmapId { get; set; }
```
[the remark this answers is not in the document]
peppy: > Additionally, after applying the above, the following:

This might have changed since when you reviewed, but it doesn't look like changes should be applied here.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/MigratePlaylistScoresToSoloScoresCommand.cs
```diff
@@ -0,0 +1,199 @@
...
+
+                        MultiplayerScore[] scores = (await db.QueryAsync<MultiplayerScore>($"SELECT * FROM multiplayer_scores WHERE playlist_item_id = {item.id}")).ToArray();
+
+                        foreach (var score in scores)
+                        {
+                            Console.WriteLine($"Reprocessing score {score.id} from playlist {item.id}");
+
+                            Ruleset ruleset = LegacyRulesetHelper.GetRulesetFromLegacyId(item.ruleset_id);
+                            HitResult maxRulesetJudgement = ruleset.GetHitResults().First().result;
+                            Dictionary<HitResult, int> statistics = JsonConvert.DeserializeObject<Dictionary<HitResult, int>>(score.statistics)!;
```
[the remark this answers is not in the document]
peppy: Has been applied.

peppy:
This should be ready for a re-review. I've tested against production (without running the actual inserts) and things look to be correct.

bdach:
Aside from https://github.com/ppy/osu-queue-score-statistics/pull/154/commits/6c8918bc7cf9ddae66ad62df343f2050c9167151, is good :+1:

bdach (approved):

PR #156, opened 2023-09-19
Don't run certain processors for failed scores

smoogipoo:
See task list item in https://github.com/ppy/osu-infrastructure/issues/19

~~Not sure if any other processors need to be adjusted - just touching the PP processor for now.~~

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor.Tests/DatabaseTest.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor.Tests/DatabaseTest.cs
@@ -12,7 +12,6 @@ using osu.Framework.Extensions.ExceptionExtensions;
 using osu.Game.Beatmaps;
 using osu.Game.Online.API.Requests.Responses;
 using osu.Game.Rulesets.Difficulty;
-using osu.Game.Rulesets.Osu.Difficulty;
 using osu.Game.Rulesets.Scoring;
 using osu.Game.Scoring;
 using osu.Server.Queues.ScoreStatisticsProcessor.Models;
@@ -172,23 +171,13 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Tests
             return beatmap;
         }
 
-        protected void AddBeatmapAttributes(Action<DifficultyAttributes>? setup = null)
+        protected void AddBeatmapAttributes<TDifficultyAttributes>(uint? beatmapId = null, Action<TDifficultyAttributes>? setup = null)
+            where TDifficultyAttributes : DifficultyAttributes, new()
         {
-            var attribs = new OsuDifficultyAttributes
+            var attribs = new TDifficultyAttributes
             {
                 StarRating = 5,
                 MaxCombo = 5,
-                AimDifficulty = 5,
-                SpeedDifficulty = 5,
-                SpeedNoteCount = 5,
-                FlashlightDifficulty = 5,
-                SliderFactor = 5,
-                ApproachRate = 5,
-                OverallDifficulty = 5,
-                DrainRate = 5,
-                HitCircleCount = 5,
-                SliderCount = 5,
-                SpinnerCount = 5
             };
 
             setup?.Invoke(attribs);
@@ -199,7 +188,7 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Tests
                 {
                     db.Insert(new BeatmapDifficultyAttribute
                     {
-                        beatmap_id = TEST_BEATMAP_ID,
+                        beatmap_id = beatmapId ?? TEST_BEATMAP_ID,
                         mode = 0,
                         mods = 0,
                         attrib_id = (ushort)a.attributeId,
--- a/osu.Server.Queues.ScoreStatisticsProcessor.Tests/DatabaseTest.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor.Tests/DatabaseTest.cs
@@ -11,6 +11,8 @@ using MySqlConnector;
 using osu.Framework.Extensions.ExceptionExtensions;
 using osu.Game.Beatmaps;
 using osu.Game.Online.API.Requests.Responses;
+using osu.Game.Rulesets.Difficulty;
+using osu.Game.Rulesets.Osu.Difficulty;
 using osu.Game.Rulesets.Scoring;
 using osu.Game.Scoring;
 using osu.Server.Queues.ScoreStatisticsProcessor.Models;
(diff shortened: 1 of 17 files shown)


bdach:
Looking at the other existing processors in the project, the max combo processor looks like the most likely candidate of one that may need a `!score.Passed` guard. Maybe the medals one too, although I'm not 100% sure if there are any funky existing medals that can be triggered by a failed score specifically. Doesn't seem like it.

osu.Server.Queues.ScoreStatisticsProcessor.Tests/DatabaseTest.cs
```diff
@@ -170,6 +172,43 @@ protected Beatmap AddBeatmap(Action<Beatmap>? beatmapSetup = null, Action<Beatma
             return beatmap;
         }
 
+        protected void AddBeatmapAttributes(Action<DifficultyAttributes>? setup = null)
```
bdach, line 175: Maybe this is gonna be nitpicking but for a helper extracted this high it doesn't seem to be terribly usable in conjunction with everything else.

- (Most relevant) Hardcodes `TEST_BEATMAP_ID` even though other tests allow the id to be changed already
- (Less relevant) Hardcodes osu! difficulty attributes

Honestly I'm not sure the method needs to live that high up, I'd move it back to `PerformanceProcessorTests` as a local helper until actually needed elsewhere. That said If I were to suggest how I'd see this as a more general helper, it'd probably be something along the lines of

```diff
diff --git a/osu.Server.Queues.ScoreStatisticsProcessor.Tests/DatabaseTest.cs b/osu.Server.Queues.ScoreStatisticsProcessor.Tests/DatabaseTest.cs
index aaf29b5..7be5a50 100644
--- a/osu.Server.Queues.ScoreStatisticsProcessor.Tests/DatabaseTest.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor.Tests/DatabaseTest.cs
@@ -12,7 +12,6 @@ using osu.Framework.Extensions.ExceptionExtensions;
 using osu.Game.Beatmaps;
 using osu.Game.Online.API.Requests.Responses;
 using osu.Game.Rulesets.Difficulty;
-using osu.Game.Rulesets.Osu.Difficulty;
 using osu.Game.Rulesets.Scoring;
 using osu.Game.Scoring;
 using osu.Server.Queues.ScoreStatisticsProcessor.Models;
@@ -172,23 +171,13 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Tests
             return beatmap;
         }
 
-        protected void AddBeatmapAttributes(Action<DifficultyAttributes>? setup = null)
+        protected void AddBeatmapAttributes<TDifficultyAttributes>(uint? beatmapId = null, Action<TDifficultyAttributes>? setup = null)
+            where TDifficultyAttributes : DifficultyAttributes, new()
         {
-            var attribs = new OsuDifficultyAttributes
+            var attribs = new TDifficultyAttributes
             {
                 StarRating = 5,
                 MaxCombo = 5,
-                AimDifficulty = 5,
-                SpeedDifficulty = 5,
-                SpeedNoteCount = 5,
-                FlashlightDifficulty = 5,
-                SliderFactor = 5,
-                ApproachRate = 5,
-                OverallDifficulty = 5,
-                DrainRate = 5,
-                HitCircleCount = 5,
-                SliderCount = 5,
-                SpinnerCount = 5
             };
 
             setup?.Invoke(attribs);
@@ -199,7 +188,7 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Tests
                 {
                     db.Insert(new BeatmapDifficultyAttribute
                     {
-                        beatmap_id = TEST_BEATMAP_ID,
+                        beatmap_id = beatmapId ?? TEST_BEATMAP_ID,
                         mode = 0,
                         mods = 0,
                         attrib_id = (ushort)a.attributeId,
diff --git a/osu.Server.Queues.ScoreStatisticsProcessor.Tests/PerformanceProcessorTests.cs b/osu.Server.Queues.ScoreStatisticsProcessor.Tests/PerformanceProcessorTests.cs
index 00eff66..ba0b66b 100644
--- a/osu.Server.Queues.ScoreStatisticsProcessor.Tests/PerformanceProcessorTests.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor.Tests/PerformanceProcessorTests.cs
@@ -6,6 +6,7 @@ using osu.Framework.Extensions.TypeExtensions;
 using osu.Game.Rulesets.Catch.Mods;
 using osu.Game.Rulesets.Mania.Mods;
 using osu.Game.Rulesets.Mods;
+using osu.Game.Rulesets.Osu.Difficulty;
 using osu.Game.Rulesets.Osu.Mods;
 using osu.Game.Rulesets.Scoring;
 using osu.Game.Rulesets.Taiko.Mods;
@@ -29,7 +30,7 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Tests
         public void PerformanceIndexUpdates()
         {
             AddBeatmap();
-            AddBeatmapAttributes();
+            AddBeatmapAttributes<OsuDifficultyAttributes>();
 
             SetScoreForBeatmap(TEST_BEATMAP_ID, score =>
             {
@@ -192,7 +193,7 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Tests
         public void FailedScoreDoesNotProcess()
         {
             AddBeatmap();
-            AddBeatmapAttributes();
+            AddBeatmapAttributes<OsuDifficultyAttributes>();
 
             ScoreItem score = SetScoreForBeatmap(TEST_BEATMAP_ID, score =>
             {

```

peppy:
I started reviewing this and was wondering if we wanted to move this into a property flag on the processor model itself, since it's so common (and probably force implementation using `abstract bool`). But maybe it's too much if we're not adding processors that often.

osu.Server.Queues.ScoreStatisticsProcessor.Tests/DatabaseTest.cs
smoogipoo: I thought there was a reason why I specified all the other values - perhaps they're required for PP to be non-zero, But it's cool that this works as is.

smoogipoo:
Updated this PR to also prevent processing other processors for failed scores.

| processor | passed scores? | failed scores? |
| --- | --- | --- |
| `HitStatisticsProcessor` | :white_check_mark: | |
| `MaxComboProcessor` | :white_check_mark: | |
| `MedalProcessor` | :white_check_mark: | :white_check_mark: [^1] |
| `PlayCountProcessor` | :white_check_mark: | :white_check_mark: |
| `PlayTimeProcessor` | :white_check_mark: | :white_check_mark: |
| `ScorePerformanceProcessor` | :white_check_mark: | |
| `TotalScoreProcessor` | :white_check_mark: | |
| `UserRankCountProcessor` | :white_check_mark: | |
| `UserTotalPerformanceProcessor` | :white_check_mark: | |

[^1]: Handled by each `IMedalAwarder`, currently false for `PackMedalAwarder`.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/IMedalAwarder.cs
```diff
@@ -15,6 +15,11 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Processors
     [UsedImplicitly]
     public interface IMedalAwarder
     {
+        /// <summary>
+        /// Whether this awarder should be run on failed scores.
+        /// </summary>
+        bool RunOnFailedScores { get; }
```
peppy, line 21: I'm not sure we'll ever use this, but no harm I guess?

osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserTotalPerformanceProcessor.cs
```diff
@@ -107,6 +109,10 @@ public async Task UpdateUserStatsAsync(UserStats userStats, int rulesetId, MySql
                 if (s.pp == null)
                     return true;
 
+                // Score must be a pass (safeguard - should be redundant with preserve flag).
```
peppy, line 112: Is this assuming that all fail scores will be `preserve==0`?

peppy (approved):
Seems fine.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserTotalPerformanceProcessor.cs
smoogipoo: That is the assumption, yes. But thinking about it it may be possible to pin those scores in the future/now?

Maybe this should’ve been a filter in the query…

peppy: nah pinning failed scores likely won't be allowed. they wouldn't have replays and feels pretty pointless

Issue #134, opened 2023-06-13
Add support for increasing ranked score


peppy:
Likely using "classic" score for community compatibility.

respektive:
I'd like to mention, that the total score statistic should also be changed to using classic scoring then, just so this isn't potentially forgotten and making the values not match and messed up. :sweat_smile:

peppy:
We're aware, yes.

PR #157, opened 2023-09-28
Use classic score in total score processor

bdach:
As per https://github.com/ppy/osu-queue-score-statistics/issues/134#issuecomment-1686230928

- [x] Depends on https://github.com/ppy/osu/pull/24957
- [x] Depends on https://github.com/ppy/osu-queue-score-statistics/pull/155 for mergeability
- [x] Depends on https://github.com/ppy/osu/pull/24924 for adjusted test values to be correct (dependency can be removed if need be).
- [x] Depends on nuget release of `osu.Game`

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor.Tests/DatabaseTest.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor.Tests/DatabaseTest.cs
@@ -132,6 +132,11 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Tests
                     { HitResult.Perfect, 5 },
                     { HitResult.LargeBonus, 0 }
                 },
+                MaximumStatistics =
+                {
+                    { HitResult.Perfect, 5 },
+                    { HitResult.LargeBonus, 2 }
+                },
                 Passed = true
             };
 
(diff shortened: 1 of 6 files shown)


peppy (approved):

smoogipoo:
Should we not be increasing `ScoreStatisticsQueueProcessor.VERSION` as a result of this? And shouldn't this support revert of both versions? Though I believe it is the first time VERSION is being used in this way, I can't imagine any other purpose for it.

peppy:
Nah, I won't be running any kind of revert operation on already applied scores. Is what it is at this point.

smoogipoo:
As long as you understand the implication is not only that you will not be doing it, but that if you ever decide to do it it will become extremely difficult to do and blindly doing it will cause those users to have negative total scores.

bdach:
> Edit: Actually it already has a check for `previousVersion >= 2`, so it's not the first time this is being done.

That's the typical "this processor did not exist on prior versions" check, isn't it? Slightly different.

> As long as you understand the implication is not only that you will not be doing it, but that if you ever decide to do it it will become extremely difficult to do and blindly doing it will cause those users to have negative total scores.

I mean... if we're bringing this up as a concern, then any change to any scoring algorithm is potentially a concern. If the scoring algorithm silently changes on the game side and then the processor version gets incremented, and the processors begin to reprocess scores, everything that touches score is now potentially not reverting what it originally applied. For this to work in a 100% controlled manner you would need a score versioning scheme etc. Which I will do, if you think that's best, I'm just not sure that it is.

smoogipoo:
It's exactly the same thing at the end of the day.

```csharp
switch (previousVersion)
{
    case < 2:
        return; // Processor didn't exist before this version.
    case < 7:
        totalScore -= standardisedScore;
        break;
    default:
        totalScore -= classicScore;
        break;
}
```

Anywhoo... https://github.com/ppy/osu-queue-score-statistics/pull/159

peppy:
With added context and IRL discussion, I think the above is a valid concern (should increase `VERSION` and add per-version revert logic).

PR #158, opened 2023-10-03
Add support for increasing ranked score

bdach:
- Closes https://github.com/ppy/osu-queue-score-statistics/issues/134
- [x] Depends on https://github.com/ppy/osu-queue-score-statistics/pull/157 (for mergeability, mostly - both pulls share prerequisites)
- [x] Depends on https://github.com/ppy/osu-queue-score-statistics/pull/160

Not much to say here really. A lot of this implementation is copy-paste from the rank count processor, with the exception of having to check the beatmap status. [Wiki says](https://osu.ppy.sh/wiki/en/Gameplay/Score/Ranked_score) ranked, loved, and approved count; that information comes from https://github.com/ppy/osu-wiki/pull/8488, which had @peppy reviewing, so seems legit.

The other part I wasn't 100% on here when querying the beatmap status, is whether to go for a local query or reuse `BeatmapStore` instead to leverage caching. Ended up doing a local query in the end as it seemed less overkill, but can switch back to the cache if you decide that's better.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Processors/RankedScoreProcessor.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Processors/RankedScoreProcessor.cs
@@ -2,13 +2,12 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System.Diagnostics;
-using Dapper;
 using JetBrains.Annotations;
 using MySqlConnector;
-using osu.Game.Beatmaps;
 using osu.Game.Online.API.Requests.Responses;
 using osu.Game.Rulesets.Scoring;
 using osu.Game.Scoring.Legacy;
+using osu.Server.Queues.ScoreStatisticsProcessor.Helpers;
 using osu.Server.Queues.ScoreStatisticsProcessor.Models;
 
 namespace osu.Server.Queues.ScoreStatisticsProcessor.Processors
@@ -23,7 +22,7 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Processors
             if (!score.Passed)
                 return;
 
-            if (!isBeatmapValidForRankedScore(score.BeatmapID, conn, transaction))
+            if (!DatabaseHelper.IsBeatmapValidForRankedCounts(score.BeatmapID, conn, transaction))
                 return;
 
             if (previousVersion >= 8)
@@ -31,7 +30,7 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Processors
                 // It is assumed that in the case of a revert, either the score is deleted, or a reapplication will immediately follow.
 
                 // First, see if the score we're reverting is the user's best (and as such included in the total ranked score).
-                var bestScore = getBestScore(score, conn, transaction);
+                var bestScore = DatabaseHelper.GetUserBestScoreFor(score, conn, transaction);
 
                 // If this score isn't the user's best on the beatmap, nothing needs to be reverted.
                 if (bestScore?.ID != score.ID)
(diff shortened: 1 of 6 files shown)


peppy:
> Not much to say here really. A lot of this implementation is copy-paste from the rank count processor, with the exception of having to check the beatmap status

Hmm, this seems weird though. Both should likely have the same checks. Maybe that was an oversight?

For what it's worth, legacy score submission has this check implemented as:

`$ranked = $approvedStatus > 0;`

Which suggests that qualified beatmaps also count. I'm not sure they should as I don't think we correctly handle rollbacks on disqualifications, so maybe I should change it to match the wiki and this PR.

bdach:
> Hmm, this seems weird though. Both should likely have the same checks. Maybe that was an oversight?

Possibly. I couldn't find any reference on the rank counts, so I just did whatever.

If they should both have the same status checks, then I'll add them to the rank count processor too.

peppy:
Yeah, let's go with that. Probably share code via some means too.

bdach:
> Yeah, let's go with that.

I've done that in this pull to save on ceremony; see test coverage in https://github.com/ppy/osu-queue-score-statistics/pull/158/commits/9f5ee2990814de843934d1e3bc6bc937346d0756 and the actual fix in https://github.com/ppy/osu-queue-score-statistics/pull/158/commits/7b487aa8f242ece8bdc4cc9a03d421f6e6c54fb0.

Notably because the user rank count processor hasn't seen production yet, I'm not bothering to consider any reverts of existing stats or anything of that sort.

> Probably share code for the database queries via some means too.

See what you think of https://github.com/ppy/osu-queue-score-statistics/pull/158/commits/f27f4a2f551d743b01373b9c74c84b26ea526d1f. Decided not to overthink it and put two helpers in an existing helper class, although it's not the most thematically-fitting thing ever.

bdach:
Blocking pending resolution of https://github.com/ppy/osu-queue-score-statistics/pull/157#issuecomment-1748147709

bdach:
Unblocking, but this depends on https://github.com/ppy/osu-queue-score-statistics/pull/160 now.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/RankedScoreProcessor.cs
```diff
@@ -0,0 +1,89 @@
...
+    public class RankedScoreProcessor : IProcessor
+    {
+        public bool RunOnFailedScores => false;
+
+        public void RevertFromUserStats(SoloScoreInfo score, UserStats userStats, int previousVersion, MySqlConnection conn, MySqlTransaction transaction)
+        {
+            if (!score.Passed)
+                return;
+
+            if (!DatabaseHelper.IsBeatmapValidForRankedCounts(score.BeatmapID, conn, transaction))
```
peppy, line 25: One interesting thing here is that with these checks, we can't use revert for the case of beatmap unrank / deletion. I'm not sure we'd want to, but felt it worth mentioning.

peppy (approved):

osu.Server.Queues.ScoreStatisticsProcessor/Processors/RankedScoreProcessor.cs
bdach: Yeah I'm not sure. I'm not very well acquainted with those processes so I don't feel like I can suggest anything at the time, but if you think it important I can spend some time finding out how that works so that I can propose alternatives.

bdach:
> I haven't checked query execution plans.

The one heavy query (finding first and second-best scores) is unchanged from the previous PR, so the hope is that the indexing improvements carry over. The other query this uses (looking up beatmap status) is by primary key so I don't expect that to be slow individually. The volume may be, not sure.

PR #162, opened 2023-10-09
Migration attempt 2

peppy:
Already running on production

- Add support for migrating all playlists to new table structure
- Move insert command preparation outside of loop for better performance
- Update `scores_high` table with link IDs
- Allow cancellation at more opportunities
- Use more efficient lookup for query


bdach:
Made a small stylistic change (https://github.com/ppy/osu-queue-score-statistics/pull/162/commits/10e48b7dfc5f5709aca130210517064abd8a162d) - looks good otherwise.

bdach (approved):

PR #163, opened 2023-10-20
Explicitly specify NC/DC as mods valid for PP

smoogipoo:
As of https://github.com/ppy/osu/pull/24640, NC no longer inherits DT and DC no longer inherits HT.

This can be merged as is without an update to nuget packages.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Processors/ScorePerformanceProcessor.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Processors/ScorePerformanceProcessor.cs
@@ -163,9 +163,23 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Processors
                     case ManiaModKey10:
                         return false;
 
+                    case ModNightcore nc:
+                        // Disallow non-default rate adjustments for now.
+                        if (!nc.SpeedChange.IsDefault)
+                            return false;
+
+                        continue;
+
                     case ModDoubleTime dt:
                         // Disallow non-default rate adjustments for now.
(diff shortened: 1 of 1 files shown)


osu.Server.Queues.ScoreStatisticsProcessor/Processors/ScorePerformanceProcessor.cs
```diff
@@ -163,13 +163,27 @@ public static bool AllModsValidForPerformance(Mod[] mods)
                     case ManiaModKey10:
                         return false;
 
+                    case ModNightcore nc:
```
peppy, line 166: wonder if it's worth casing `ModRateAdjust` at the end of the switch to cover all of these? rather than having individual handling?

smoogipoo: Doesn't it feel iffy to blindly allow all rate adjust mods?

bdach: Judging that we're in this predicament due to relying on inheritance hierarchies, I think specifying mod classes explicitly is better.

You could _maybe_ try to keep one unified copy of the code under four case labels. Maybe.

peppy: > You could maybe try to keep one unified copy of the code under four case labels. Maybe.

Might read better, not fussed either way just curious on preference.

peppy (approved):

PR #166, opened 2023-11-09
Add support for incrementing global and beatmap playcounts

peppy:
Covers `osu_counts` playcount, as well as beatmap and beatmap set level playcounts. All implemented using the same logic that as web-10.

- [x] Depends on https://github.com/ppy/osu-queue-score-statistics/pull/164

personal reminder: consider backfilling these fields from solo_scores when deploying.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor.Tests/PlayCountProcessorTests.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor.Tests/PlayCountProcessorTests.cs
@@ -22,9 +22,9 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Tests
         }
 
         [Fact]
-        public void TestGlobalPlaycountIncrement()
+        public void TestGlobalPlaycountsIncrement()
         {
-            const int attempt_count = 5000;
+            const int attempt_count = 100;
 
             AddBeatmap();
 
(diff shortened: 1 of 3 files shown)


osu.Server.Queues.ScoreStatisticsProcessor.Tests/PlayCountProcessorTests.cs
```diff
@@ -19,6 +21,27 @@ public void TestPlaycountIncreaseMania()
...
         }
 
+        [Fact]
+        public void TestGlobalPlaycountsIncrement()
+        {
+            const int attempt_count = 100;
+
+            AddBeatmap();
+
+            Task.Run(() =>
```
smoogipoo, line 31: Is there a reason why this is running in a Task that doesn't seem to be awaited anywhere? It almost looks like this task could still be running even after this test is complete, is that not the case?

Fwiw, I believe xunit tests can be `public async Task`

smoogipoo (approved):

osu.Server.Queues.ScoreStatisticsProcessor.Tests/PlayCountProcessorTests.cs
peppy: Yeah that's probably bad, will fix.

smoogipoo (approved):

PR #176, opened 2023-12-05
Fix legacy score ID column type

smoogipoo:
😶


bdach:
For reference:

```
mysql> describe solo_scores_legacy_id_map;
+--------------+-------------------+------+-----+---------+-------+
| Field        | Type              | Null | Key | Default | Extra |
+--------------+-------------------+------+-----+---------+-------+
| ruleset_id   | smallint unsigned | NO   | PRI | NULL    |       |
| old_score_id | bigint unsigned   | NO   | PRI | NULL    |       |
| score_id     | bigint unsigned   | NO   |     | NULL    |       |
+--------------+-------------------+------+-----+---------+-------+
```

bdach (approved):

PR #177, opened 2023-12-12
Fix Daycore mod not having an effect on pp values when realtime difficulty is off

bdach:
Which it currently is in production (and will continue to be in the foreseeable future).

Closes https://github.com/ppy/osu/issues/25719.


peppy (approved):
I think this is okay

smoogipoo (approved):

PR #178, opened 2023-12-12
Add support for automatic index schema lookup

peppy:
- Add automatic schema support
- Remove unnecessary envvar set


bdach (approved):
Looks correct to me

smoogipoo (approved):

PR #180, opened 2023-12-13
Disallow classic mod from giving PP (for now)

smoogipoo:
Since we're not sure yet on the implementation of the mechanics of this mod (e.g. https://github.com/ppy/osu/issues/11769), it's easiest to just not allow it to be ranked for now. Scores coming from stable are still allowed to give PP.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Processors/ScorePerformanceProcessor.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Processors/ScorePerformanceProcessor.cs
@@ -192,7 +192,7 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Processors
                         continue;
 
                     case ModClassic:
-                        // Classic mod is only allowed if it's attached to legacy scores (null build ID).
+                        // Classic mod is only allowed if it's attached to legacy scores.
                         return score.IsLegacyScore;
 
                     // The mods which are allowed.
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Processors/ScorePerformanceProcessor.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Processors/ScorePerformanceProcessor.cs
@@ -121,7 +121,7 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Processors
                 Ruleset ruleset = LegacyRulesetHelper.GetRulesetFromLegacyId(score.RulesetID);
                 Mod[] mods = score.Mods.Select(m => m.ToMod(ruleset)).ToArray();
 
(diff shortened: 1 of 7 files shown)


osu.Server.Queues.ScoreStatisticsProcessor/Processors/ScorePerformanceProcessor.cs
```diff
@@ -191,6 +191,10 @@ public static bool AllModsValidForPerformance(Mod[] mods)
 
                         continue;
 
+                    case ModClassic:
+                        // Classic mod is only allowed if it's attached to legacy scores (null build ID).
+                        return score.BuildID == null;
```
bdach, line 196: Isn't examining that `LegacyScoreID` is not null the standard / canonical way to check if a `SoloScoreInfo` is legacy or not? I worry about introducing multiple heuristics for the same thing.

smoogipoo: Hmmm...

https://github.com/ppy/osu-queue-score-statistics/blob/3def49348ed93535df38d4bc8ffdbdb94709df47/osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserTotalPerformanceProcessor.cs#L119-L124

@peppy any opinions?

bdach: Just for context, I'm referring to

https://github.com/ppy/osu-queue-score-statistics/blob/3def49348ed93535df38d4bc8ffdbdb94709df47/osu.Server.Queues.ScoreStatisticsProcessor/ScoreStatisticsQueueProcessor.cs#L101-L105
https://github.com/ppy/osu-queue-score-statistics/blob/3def49348ed93535df38d4bc8ffdbdb94709df47/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/DeleteImportedHighScoresCommand.cs#L78-L79

as places that use the other method.

smoogipoo: Will wait on a new osu! nuget package before fixing this.

bdach:
I have attempted to fix test failures here but I am not sure everyone will agree as to what I did, so please scrutinise https://github.com/ppy/osu-queue-score-statistics/pull/180/commits/b9b51aceb9075767a91a774dbfafa2b6f852468f especially at your discretion.

Rest seems fine to me.

bdach (approved):

smoogipoo:
Looks fine to me

peppy (approved):

PR #181, opened 2023-12-13
Allow limited processing of imported high scores

bdach:
Aims to resolve https://github.com/ppy/osu-infrastructure/issues/25.

This PR allows a limited subset of PP-related processors to run for legacy scores, and changes the score import job to push imported scores for aforementioned processing when running in watch/indefinite mode.

I am very intentionally opening this as draft to first ensure that I am in the correct wheelhouse when it comes to what the issue above intended to convey. **Therefore, this is extremely untested code. It passes tests, but that means nothing for the import job. I have not attempted to run the import job. It only works in my head. Please give this a quick general review, but please do not merge this until I have tested it.**

I **will** attempt to run it and test it, to the best of my ability, using ad-hoc means, namely manual database inserts - these are the only means I have, given that stable and surrounding infra is a closed-off system - when I receive confirmation that this looks roughly correct when it comes to the general shape and meets the general expectations. I'll also add extended test coverage for the pp processors, ensuring that they do in fact process legacy scores.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/ImportHighScoresCommand.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/ImportHighScoresCommand.cs
@@ -84,6 +84,7 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Queue
         private long lastLatencyCheckTimestamp;
 
         private ElasticQueuePusher? elasticQueueProcessor;
+        private ScoreStatisticsQueueProcessor? scoreStatisticsQueueProcessor;
 
         private static int currentReportInsertCount;
         private static int currentReportUpdateCount;
@@ -168,12 +169,26 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Queue
                 }
             }
 
+            Console.WriteLine();
+            Console.WriteLine($"Sourcing from {highScoreTable} for {ruleset.ShortName} starting from {lastId}");
+            Console.WriteLine($"Inserting into scores and processing {(watchMode ? "indefinitely" : "as single run")}");
+
             if (watchMode)
             {
-                // Assume that `score_process_queue` is processed by a single process sequentially.
-                // We don't want to import until the score has finished processing or we may import incorrect pp data.
+                scoreStatisticsQueueProcessor = new ScoreStatisticsQueueProcessor();
+
+                if (SkipIndexing)
+                {
+                    throw new NotSupportedException("Skipping indexing in watch mode is not supported, "
+                                                    + $"as watch mode will push imported scores to the {scoreStatisticsQueueProcessor.QueueName} redis queue in order to process PP, "
+                                                    + "and once that score is processed on the aforementioned queue, it will be pushed for indexing to ES.");
+                }
+
+                Console.WriteLine($"Pushing imported scores to redis queue {scoreStatisticsQueueProcessor.QueueName}");
+
+                // We are going to enqueue the score for pp processing ourselves, so we don't care about its current processing status on the old queue.
                 maxProcessableId = ulong.MaxValue;
-                where = "WHERE score_id >= @lastId AND (SELECT score_id FROM score_process_queue WHERE score_id = h.score_id AND status > 0 AND mode = @rulesetId LIMIT 1) IS NOT NULL";
+                where = "WHERE score_id >= @lastId";
             }
             else
             {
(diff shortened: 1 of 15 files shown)


peppy:
Direction looks correct, if we ignore the conflicts with the other PR 😅 .

bdach:
Have rebased (fixing no less than two back-to-front conditionals along the way), but haven't tested properly yet. Hopefully will get that done tomorrow.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/ImportHighScoresCommand.cs
```diff
@@ -168,12 +169,26 @@ public async Task<int> OnExecuteAsync(CancellationToken cancellationToken)
...
+
             if (watchMode)
             {
-                // Assume that `score_process_queue` is processed by a single process sequentially.
-                // We don't want to import until the score has finished processing or we may import incorrect pp data.
+                scoreStatisticsQueueProcessor = new ScoreStatisticsQueueProcessor();
+
+                if (SkipIndexing)
+                {
+                    throw new NotSupportedException("Skipping indexing in watch mode is not supported, "
```
peppy, line 187: Hmm, this feels a bit weird. To confirm, we're *always* skipping indexing in watch mode because we delegate that to `osu-queue-score-statistics` right? Might be worth mentioning that either in an inline comment or in the exception message because right now it's not too clear (feels like it's saying "you can't skip indexing in this mode").

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/ImportHighScoresCommand.cs
```diff
@@ -306,22 +317,25 @@ public async Task<int> OnExecuteAsync(CancellationToken cancellationToken)
                         return -1;
                     }
 
-                    if (elasticQueueProcessor != null)
+                    if (scoreStatisticsQueueProcessor != null)
```
peppy, line 325: As far as I can tell, we should always be pushing to either one queue or the other at this point, right?

In which case, I'd like that to be asserted, and probably use an `if-else` here instead of two `if`s.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/ImportHighScoresCommand.cs
bdach: I mean... yes the import job itself will not index, but the resulting score processing of the imported score _will_ index. So the net result of the operation will be that _an_ indexing _will_ happen. While the intent of the flag seemed that _no_ indexing should take place, not now, not later. So, also yes, the thing was supposed to be a "stop, this will do things you may not expect it to do".

If that is weird, or I misread intent with respect to the flag, I can remove this.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/ImportHighScoresCommand.cs
bdach: I pushed https://github.com/ppy/osu-queue-score-statistics/pull/181/commits/0817a9a51d93274ba6823893287531c5ed6251e9 to attempt to address this. Not sure if it's going to be to your liking?

bdach:
So I began trying to test this. Didn't get very far, because of other screw-ups (tl;dr: I accidentally nuked my local osu-web setup's db by running this project's test suite on it, as it truncates `osu_beatmaps` among others), but there is one apparent possible issue.

To be sure whether it is an issue, I would need an answer to a few questions that requires implicit knowledge, and those are:
- What inserts rows to `score_process_queue`?
- Will whatever inserts these rows continue to insert said rows after `osu-performance` is shut down?

Because if it will not, then there are several checks in the import job for watch mode that rely on `score_process_queue` to be able to tell where to pick up from. And I'm not sure what to do with those. The answer is probably delete them?

When I attempted to run the job, the following happened:

- I ran `queue import-high-scores` without ever having ran anything before. It correctly recognised that I was "bootstrapping a local dev setup" - after I fixed a defective check (see https://github.com/ppy/osu-queue-score-statistics/pull/181/commits/353795f840098d7ef1713b6d14b4facca2fac444) - and proceeded to import all seeded high scores from start to end. This is good.
- Then I ran `queue import-high-scores` again, hoping to enter watch mode proper. But I was not attempting to even simulate `score_process_queue` insertions in any way, so the job again assumed I had never done anything before and was trying to bootstrap, and thus tried to import everything again, which seems... undesirable?

And so I'm not super sure where I'm going wrong here, but something isn't quite right. Either I'm not manually simulating stable submission properly, or there is an actual issue with the code.

peppy:
> What inserts rows to score_process_queue?

Score submitting, web-10, which will continue happening into the foreseeable future.

> Then I ran queue import-high-scores again, hoping to enter watch mode proper. But I was not attempting to even simulate score_process_queue insertions in any way, so the job again assumed I had never done anything before and was trying to bootstrap, and thus tried to import everything again, which seems... undesirable?

You'll need to insert into the `score_process_queue` table yeah. If there's no rows it assumes there's no ongoing processing. This is expected behaviour since it's guaranteed to have rows on production.

bdach:
> You'll need to insert into the `score_process_queue` table yeah. If there's no rows it assumes there's no ongoing processing. This is expected behaviour since it's guaranteed to have rows on production.

Right. `osu-performance` [updates this table to mark completion of processing](https://github.com/ppy/osu-performance/blob/3d403354321c5e0cd0cd9c9cf587d5b88a0ba78f/src/performance/Score.cpp#L51). So I assume if this job is to eventually be the replacement for `osu-performance`, then I presume I will need to mimic that too? (maybe behind a flag?)

peppy:
If we're back-populating the old table then that will be the case, yes.

An alternative which may sit better is leaving `osu-performance` running and updating `osu_scores_*_high` but not updating user totals.

smoogipoo:
Haven't tested it, but have done the above in https://github.com/ppy/osu-performance/pull/177

peppy:
I'm going to merge this as it's too hard to maintain multiple code changes on this single file with these changes pending any longer.

peppy (approved):

PR #182, opened 2023-12-14
Fix oversights in processing logic

peppy:
The previous implementation of using `score_process_queue` to more safely query scores was flawed in multiple ways. Explaining this part is hard, so if the code doesn't speak for itself, I'll just merge this.

This also adds more sanity to which mode the import task  chooses to enter, including bailing if attempting to run in a completely incorrect state.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/ImportHighScoresCommand.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/ImportHighScoresCommand.cs
@@ -173,7 +173,7 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Queue
                 // Assume that `score_process_queue` is processed by a single process sequentially.
                 // We don't want to import until the score has finished processing or we may import incorrect pp data.
                 maxProcessableId = ulong.MaxValue;
-                where = "LEFT JOIN score_process_queue USING (score_id) WHERE score_id >= @lastId AND status > 0 AND mode = @rulesetId";
+                where = "WHERE score_id >= @lastId AND (SELECT score_id FROM score_process_queue WHERE score_id = h.score_id AND status > 0 AND mode = @rulesetId LIMIT 1) IS NOT NULL";
             }
             else
             {
@@ -204,7 +204,7 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Queue
                     if (CheckSlaveLatency)
                         checkSlaveLatency(dbMainQuery);
 
-                    var highScores = await dbMainQuery.QueryAsync<HighScore>($"SELECT * FROM {highScoreTable} {where}" +
+                    var highScores = await dbMainQuery.QueryAsync<HighScore>($"SELECT * FROM {highScoreTable} h {where}" +
                                                                              " ORDER BY score_id LIMIT @scoresPerQuery", new
                     {
                         lastId,
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/ImportHighScoresCommand.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/ImportHighScoresCommand.cs
@@ -136,17 +136,32 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Queue
             DateTimeOffset start = DateTimeOffset.Now;
 
             ulong lastId;
-
+            string where;
             bool singleRun = StartId.HasValue;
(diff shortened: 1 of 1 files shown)


osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/ImportHighScoresCommand.cs
```diff
@@ -347,7 +366,7 @@ private ulong getMaxProcessable(Ruleset ruleset)
                 // when doing a single run, we need to make sure not to run into scores which are in the process queue (to avoid
                 // touching them while they are still being written).
                 using (var db = DatabaseAccess.GetConnection())
-                    return db.QuerySingle<ulong?>($"SELECT MIN(score_id) FROM score_process_queue WHERE is_deleted = 0 AND mode = {ruleset.RulesetInfo.OnlineID}") - 1 ?? ulong.MaxValue;
+                    return db.QuerySingle<ulong?>($"SELECT MIN(score_id) FROM score_process_queue WHERE is_deletion = 0 AND mode = {ruleset.RulesetInfo.OnlineID}") - 1 ?? ulong.MaxValue;
```
bdach, line 369: My only doubt is that neither `is_deleted` nor `is_deletion` exist in my local database, so I'm unable to check which one is the correct if anything, but I suppose I'll just presume this is correct...

bdach (approved):

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/ImportHighScoresCommand.cs
```diff
@@ -135,49 +135,67 @@ public async Task<int> OnExecuteAsync(CancellationToken cancellationToken)
...
+            string where;
+            bool watchMode = !StartId.HasValue;
+            ulong maxProcessableId;
+
+            if (watchMode)
+            {
+                using var db = DatabaseAccess.GetConnection();
+
+                ulong? lastImportedLegacyScore = db.QuerySingleOrDefault<ulong?>($"SELECT MAX(old_score_id) FROM score_legacy_id_map WHERE ruleset_id = {RulesetId}") ?? 0;
+                ulong? lowestProcessQueueEntry = db.QuerySingle<ulong?>($"SELECT MIN(score_id) FROM score_process_queue WHERE is_deletion = 0 AND mode = {ruleset.RulesetInfo.OnlineID}") - 1 ?? ulong.MaxValue;
```
smoogipoo, line 148: I hope you've actually tested this with a "local dev setup", yeah? Because this like looks awfully weird - that it's a `QuerySingle` and not a `QuerySingleOrDefault` which looks to be more appropriate.

peppy: I think you're correct here, will update.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/ImportHighScoresCommand.cs
peppy: Here's the new table structure. I think I posed in #infra as needing an update in the schema.

```sql
CREATE TABLE `score_process_queue` (
  `queue_id` int unsigned NOT NULL AUTO_INCREMENT,
  `score_id` bigint unsigned NOT NULL,
  `mode` tinyint unsigned NOT NULL,
  `start_time` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
  `update_time` timestamp NULL DEFAULT NULL ON UPDATE CURRENT_TIMESTAMP,
  `status` tinyint(1) NOT NULL DEFAULT '0',
  `is_deletion` tinyint(1) NOT NULL DEFAULT '1' COMMENT 'marks not just deletion but basically anything that isn''t a new score',
  PRIMARY KEY (`queue_id`,`start_time`),
  KEY `status` (`status`),
  KEY `lookup_v3` (`mode`,`status`,`queue_id`),
  KEY `temp_pp_processor` (`score_id`)
) ENGINE=InnoDB AUTO_INCREMENT=2942130695 DEFAULT CHARSET=utf8mb3
```

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/ImportHighScoresCommand.cs
peppy: https://github.com/peppy/osu-queue-score-statistics/commit/6f7382e3d314cc3d7df848fc5468ec1c9b2c4358k

PR #184, opened 2023-12-18
Update table names

peppy:
Basically locking in the current live build we have, and moving ahead towards the next release, so let's get this in `master`

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor.Tests/DatabaseTest.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor.Tests/DatabaseTest.cs
@@ -61,7 +61,7 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Tests
                 try
                 {
                     // Until osu-web adds the alias views, let's take things into our own hands.
-                    var _ = db.Query<ulong>("SELECT MIN(id) FROM scores");
+                    var _ = db.Query<ulong?>("SELECT MIN(id) FROM scores");
                 }
                 catch
                 {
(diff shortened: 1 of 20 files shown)


osu.Server.Queues.ScoreStatisticsProcessor.Tests/DatabaseTest.cs
```diff
@@ -58,17 +58,32 @@ protected DatabaseTest()
                 if (db.QueryFirstOrDefault<int?>("SELECT * FROM osu_counts WHERE name = 'is_production'") != null)
                     throw new InvalidOperationException("You are trying to do something very silly.");
 
+                try
+                {
+                    // Until osu-web adds the alias views, let's take things into our own hands.
```
nanaya, line 63: osu-web already added the [aliases](https://github.com/ppy/osu-web/blob/master/database/migrations/2023_10_31_074205_create_score_table_aliases.php)

peppy: Yeah, I was led astray from the error message when running `TRUNCATE` (table doesn't exist). Have removed this.

bdach:
There are a few usages of `multiplayer_score_links` remaining:

https://github.com/ppy/osu-queue-score-statistics/blob/9ee3b673a29fd3acc074241565bfe3866fec4bb8/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/MigratePlaylistScoresToSoloScoresCommand.cs#L47-L48
https://github.com/ppy/osu-queue-score-statistics/blob/master/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/MigratePlaylistScoresToSoloScoresCommand.cs#L152

but that may not just be a rename as the table didn't only change name to `multiplayer_playlist_item_scores`, it also got considerably narrower:

```
mysql> describe multiplayer_playlist_item_scores;
+------------------+-----------------+------+-----+---------+-------+
| Field            | Type            | Null | Key | Default | Extra |
+------------------+-----------------+------+-----+---------+-------+
| user_id          | int unsigned    | NO   |     | NULL    |       |
| playlist_item_id | bigint unsigned | NO   |     | NULL    |       |
| score_id         | bigint unsigned | NO   |     | NULL    |       |
+------------------+-----------------+------+-----+---------+-------+
3 rows in set (0.00 sec)
```

Not sure whether that's something to fix in here...

smoogipoo:
Running this across the spreadsheet generator just to make sure all the processes work correctly: https://github.com/ppy/osu/actions/runs/7256587606

smoogipoo:
I'll merge this as is. As I understand it that was a once-off command and can be fixed when/if it's ever needed again.

smoogipoo (approved):

PR #187, opened 2023-12-20
Add more PP safeguards

smoogipoo:
- We noticed there are already scores in the database that have been set with lazer that don't have a `build_id`. I previously had a condition for this, that was changed to an assert following discussion, that actually cannot be asserted.
- Some scores present in the database may already have PP values from before we had per-mod filtering for PP (e.g. RX scores or such). These should not be considered towards the user total.


smoogipoo:
Running a diffcalc comparison (master vs this PR) https://github.com/ppy/osu/actions/runs/7271428812

bdach (approved):
Looks correct, will wait for the diffcalc run though

bdach:
[Spreadsheet](https://docs.google.com/spreadsheets/d/149pXBtOquaEkiDESmu0hOfS85HbZWsKD2-LhJRvok8Q/edit) looks clean, which seems good?

PR #190, opened 2023-12-27
Add flag to allow skipping score-statistics-processor in watch mode

peppy:
This will let me run watch mode in the short term (until we decide to switch to the new processor a bit later on).


peppy:
Is deployed and working

bdach (approved):

PR #192, opened 2023-12-29
Add new watch mode command

peppy:
Posting this for review, but I'l be testing this live immediately.

Basically, this adds a new watch mode command. The old one still exists and can still watch, but the new one handles things in a slightly better way:

- The source of new scores is now tailing `score_process_queue` rather than `osu_scores_high`. This was done to make us aware of deletions performed by web-10 (which are inserted into the queue table with `is_deletion = true`.
- All batching is removed to simplify things as much as possible. For the watch mode we don't need any kind of insane performance, so I cut things back to the basics.
- All slave checks and rate adjusts are removed for the same reason.
- Deleting rows is now supported by `BatchInserter`. Note that this will only take effect with the new watch mode command, as it relies on the `is_deletion` flag to be included in the provided `HighScore`s.

If all goes well here, eventually I will remove the watch mode from the old command and make it specific to running a full import.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/BatchInserter.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/BatchInserter.cs
@@ -125,8 +125,6 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Queue
                 {
                     try
                     {
-                        bool isDeletion = highScore.user_id == 0 && highScore.score == 0;
-
                         if (highScore.score_id == 0)
                         {
                             // Something really bad probably happened, abort for safety.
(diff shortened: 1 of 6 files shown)


osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/BatchInserter.cs
```diff
@@ -0,0 +1,428 @@
...
+                                deleteOldId.Value = existingMapping.old_score_id;
+                                deleteNewId.Value = existingMapping.score_id;
+
+                                if (!deleteCommand.IsPrepared)
+                                    await deleteCommand.PrepareAsync();
+
+                                await runCommand(deleteCommand);
+                                await enqueueForFurtherProcessing(existingMapping.score_id, db, transaction, true);
+
+                                // TODO: delete
```
bdach, line 188: What does this todo mean?

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/HighScore.cs
```diff
@@ -0,0 +1,37 @@
...
+        public int enabled_mods { get; set; }
+        public DateTimeOffset date { get; set; }
+        public float pp { get; set; }
+        public bool replay { get; set; }
+        public bool hidden { get; set; }
+        public string country_acronym { get; set; } = null!;
+
+        // This comes from score_process_queue. Used in join context.
+        public uint queue_id { get; set; }
+        public bool is_deletion { get; set; }
```
bdach, line 35: This feels dangerous. `ImportHighScoresCommand` uses this model without the join.

https://github.com/ppy/osu-queue-score-statistics/blob/9f117ceff7e0a8120f052700c194f21c26e132d7/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/ImportHighScoresCommand.cs#L212-L213

Could potentially lead to accidentally treating garbage data as if it wasn't garbage.

I'd probably make these nullable and assert locally in places where the join context is guaranteed to be there.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/WatchHighScoresCommand.cs
```diff
@@ -0,0 +1,228 @@
...
+{
+    /// <summary>
+    /// Watches for new high scores from the osu_scores_high tables and imports into the new scores table.
+    /// </summary>
+    /// <remarks>
+    /// This command is written under the assumption that only one importer instance is running concurrently.
+    /// This is important to guarantee that scores are inserted in the same sequential order that they originally occured,
+    /// which can be used for tie-breaker scenarios.
+    /// </remarks>
+    [Command("watch-high-scores", Description = "Imports high scores from the osu_scores_*_high tables into the new scores table.")]
```
bdach, line 25:
```suggestion
    [Command("watch-high-scores", Description = "Watches for new high scores from the osu_scores_high tables and imports into the new scores table..")]
```

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/WatchHighScoresCommand.cs
```diff
@@ -0,0 +1,228 @@
...
+    public class WatchHighScoresCommand
+    {
+        /// <summary>
+        /// The ruleset to run this import job for.
+        /// </summary>
+        [Option(CommandOptionType.SingleValue, Template = "--ruleset-id")]
+        public int RulesetId { get; set; }
+
+        /// <summary>
+        /// When set to <c>true</c> and in watch mode, scores will not be queued to the score statistics processor,
```
bdach, line 35:
```suggestion
        /// When set to <c>true</c>, scores will not be queued to the score statistics processor,
```

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/WatchHighScoresCommand.cs
```diff
@@ -0,0 +1,228 @@
...
+        public async Task<int> OnExecuteAsync(CancellationToken cancellationToken)
+        {
+            Ruleset ruleset = LegacyRulesetHelper.GetRulesetFromLegacyId(RulesetId);
+            string highScoreTable = LegacyDatabaseHelper.GetRulesetSpecifics(RulesetId).HighScoreTable;
+
+            // When running in watch mode, we need to ascertain a few things:
+            // - There is active ongoing processing in `score_process_queue` – if not, there will be nothing to watch and we will switch to full run mode.
+            // - Check whether the full run (non-watch) has caught up to recent scores (ie. processed up to a `score_id` contained in `score_process_queue`,
+            //   which generally keeps ~3 days of scores). If not, we should warn that there is a gap in processing, and an extra full run should be performed
+            //   to catch up.
```
bdach, line 68: How much of this is even relevant to this command at this point?

Notably the  "no ongoing processing" thing is reflected by an early-return below and I'm not sure that's expected behaviour from a pure watch command.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/WatchHighScoresCommand.cs
```diff
@@ -0,0 +1,228 @@
...
+                        RulesetId,
+                    })).ToArray();
+
+                    if (highScores.Length == 0)
+                    {
+                        Thread.Sleep(500);
+                        continue;
+                    }
+
+                    var inserter = new BatchInserter(ruleset, highScores, importLegacyPP: !SkipScoreProcessor, skipExisting: true, skipNew: false, dryRun: DryRun);
```
bdach, line 132: seems backwards?

```suggestion
                    var inserter = new BatchInserter(ruleset, highScores, importLegacyPP: SkipScoreProcessor, skipExisting: true, skipNew: false, dryRun: DryRun);
```

if score processor is skipped, then the implication is that the pp is already calculated by the legacy pathways, in which case it should be inserted across from the legacy tables

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/BatchInserter.cs
peppy: implement deletion, but i guess i did elsewhere

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/WatchHighScoresCommand.cs
peppy: Luckily this was not back-to-front in the commit-to-master that is deployed 😅

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/WatchHighScoresCommand.cs
peppy: Not relevant, will just nuke.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/HighScore.cs
peppy: Yeah, not sure yet. Dapper doesn't even seem to be handling this correctly.

peppy: Made nullable at least.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/WatchHighScoresCommand.cs
```diff
@@ -0,0 +1,237 @@
...
+            {
+                while (!cancellationToken.IsCancellationRequested)
+                {
+                    HighScore[] highScores = (await dbMainQuery.QueryAsync<HighScore>($"SELECT * FROM osu.score_process_queue LEFT JOIN {highScoreTable} h USING (score_id) WHERE queue_id >= @lastQueueId AND mode = {RulesetId} ORDER BY queue_id LIMIT 1000", new
+                                             {
+                                                 lastQueueId,
+                                                 RulesetId,
+                                             }))
+                                             // there might be multiple queue entries for the same insert. this can cause issues due to how we do the mapping lookup so let's fix that.
+                                             .DistinctBy(s => s.score_id)
```
bdach, line 131: The thing that worried me slightly here is I'm not sure _which_ of the duplicate queue entries this `.DistinctBy()` is going to pick. But I'm not entirely sure this matters either, because the worst case scenario of what could happen here is:

- select returns 2 rows for single score, e.g. one with `queue_id = 100000` and `queue_id = 100002`
- distinct picks one of the two arbitrarily
- the `queue_id` is later only read from the last score in a batch, to determine where to start the next batch
- so the worst case scenario here would be basically that some scores will be imported twice (because batches will partially overlap)

so I *think* it still works......

bdach (approved):
i *think* this is okay

Issue #191, opened 2023-12-28
pp shouldn't be awarded when changing mod settings (for now)


cl8n:
one of my friends noticed that they still have pp calculated when changing the Flashlight size or toggling off the option where it shrinks based on combo. for now, I think any new mod settings should also disqualify the score from pp calculation until they are weighted properly.

I believe this happens because `AllModsValidForPerformance` is currently ignoring all mod settings besides `SpeedChange` for dt/ht

PR #193, opened 2023-12-29
Disallow giving PP for mods with non-default settings

smoogipoo:
Resolves https://github.com/ppy/osu-queue-score-statistics/issues/191

Prereqs:
- [x] https://github.com/ppy/osu/pull/26216

Blanket applying for now.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor.Tests/PerformanceProcessorTests.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor.Tests/PerformanceProcessorTests.cs
@@ -293,7 +293,8 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Tests
             var mods = new Mod[]
             {
                 new OsuModDoubleTime { SpeedChange = { Value = 1.1 } },
-                new OsuModClassic { NoSliderHeadAccuracy = { Value = false } }
+                new OsuModClassic { NoSliderHeadAccuracy = { Value = false } },
+                new OsuModFlashlight { SizeMultiplier = { Value = 2 } }
             };
 
             foreach (var mod in mods)
--- a/osu.Server.Queues.ScoreStatisticsProcessor.Tests/PerformanceProcessorTests.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor.Tests/PerformanceProcessorTests.cs
@@ -287,6 +287,19 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Tests
             });
         }
 
+        [Fact]
+        public void ModsWithSettingsAreDisallowed()
+        {
+            var mods = new Mod[]
+            {
+                new OsuModDoubleTime { SpeedChange = { Value = 1.1 } },
+                new OsuModClassic { NoSliderHeadAccuracy = { Value = false } }
+            };
+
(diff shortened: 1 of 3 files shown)


bdach:
I was gonna apply a carve-out for perfect/sudden death/muted but this is probably also ok for the time being.

osu.Server.Queues.ScoreStatisticsProcessor.Tests/PerformanceProcessorTests.cs
```diff
@@ -287,6 +287,19 @@ public void LegacyScoreDoesProcess()
             });
         }
 
+        [Fact]
+        public void ModsWithSettingsAreDisallowed()
+        {
+            var mods = new Mod[]
+            {
+                new OsuModDoubleTime { SpeedChange = { Value = 1.1 } },
+                new OsuModClassic { NoSliderHeadAccuracy = { Value = false } }
```
bdach, line 296: Generally this particular assert isn't doing much in the context of settings being default or not; `OsuModClassic` is failing the `AllModsValidForPerformance()` check here because the `SoloScoreInfo` is not a legacy score.

I'd probably add another case here.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/ScorePerformanceProcessor.cs
```diff
@@ -206,18 +174,29 @@ public static bool AllModsValidForPerformance(SoloScoreInfo score, Mod[] mods)
...
                     case OsuModTouchDevice:
                     // mania-specific:
                     case ManiaKeyMod:
                     case ManiaModMirror:
-                        continue;
+                        break;
+
+                    // Classic mod is only allowed on legacy scores.
+                    case ModClassic:
+                        return score.IsLegacyScore;
```
bdach, line 191: This is perhaps a side note and not relevant to the core of this pull but I realise now - after making a mistake in my attempt at this - that this is _kinda_ wrong (although not in a way that would _functionally_ matter) and, if wanting to be 100% correct, should probably be

```suggestion
                    case ModClassic:
                        if (!score.IsLegacyScore)
                            return false;
                            
                        break;
```

because this is inside a switch *inside* a foreach. So what this ends up doing is that if a score is passed in here with classic mod and DT, for example, it won't bother to check the DT mod once it encounters the classic mod and checks that the score is legacy.

Maybe this was by design, I dunno.

osu.Server.Queues.ScoreStatisticsProcessor.Tests/PerformanceProcessorTests.cs
smoogipoo: Added a case for flashlight. Kept the classic mod case because it can't hurt.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/ScorePerformanceProcessor.cs
smoogipoo: Good catch. I don't think this matters in terms of legacy scores, but good to have.

bdach (approved):

bdach:
@peppy we probably wanna deploy this one to production before going live.

peppy (approved):

PR #194, opened 2023-12-29
Verify command

peppy:
Just scoped to pp for now due to YESTERDAY'S CHAOS.

Running at about 1m rows per minute. ~36 hours for a full process.

- [x] Depends on https://github.com/ppy/osu-queue-score-statistics/pull/192 for conflict avoidance.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/VerifyImportedScoresCommand.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/VerifyImportedScoresCommand.cs
@@ -81,8 +81,6 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Maintenance
 
                     if (importedScore.HighScore == null)
                     {
-                        Console.WriteLine($"{importedScore.id}: Original score was deleted!!");
-
                         // TODO: delete imported score
                         Interlocked.Increment(ref deleted);
                         continue;
@@ -94,12 +92,14 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Maintenance
                         Interlocked.Increment(ref fail);
 
                         await conn.ExecuteAsync($"INSERT INTO score_performance VALUES ({importedScore.id}, {importedScore.HighScore.pp})");
+                        elasticItems.Add(new ElasticQueuePusher.ElasticScoreItem { ScoreId = (long?)importedScore.id });
                         continue;
                     }
 
                     if (!check(importedScore.id, "performance", importedScore.pp, importedScore.HighScore.pp))
                     {
                         await conn.ExecuteAsync($"UPDATE score_performance SET pp = {importedScore.HighScore.pp} WHERE score_id = {importedScore.id}");
+                        elasticItems.Add(new ElasticQueuePusher.ElasticScoreItem { ScoreId = (long?)importedScore.id });
                         Interlocked.Increment(ref fail);
                     }
 
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/VerifyImportedScoresCommand.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/VerifyImportedScoresCommand.cs
@@ -1,4 +1,5 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+
 // See the LICENCE file in the repository root for full licence text.
 
 using System;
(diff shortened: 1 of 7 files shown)


osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/BatchInserter.cs
```diff
@@ -248,6 +228,32 @@ private async Task run(HighScore[] scores)
             }
         }
 
+        public static string SerialiseScore(Ruleset ruleset, HighScore highScore, ScoreInfo referenceScore)
+        {
+            var serialisedScore = JsonConvert.SerializeObject(new SoloScoreInfo
+            {
+                // id will be written below in the UPDATE call.
```
bdach, line 235:
```suggestion
```

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/VerifyImportedScoresCommand.cs
```diff
@@ -0,0 +1,180 @@
...
+            {
+                Console.WriteLine($"{scoreId}: {name} doesn't match ({imported} vs {original})");
+                return false;
+            }
+
+            return true;
+        }
+
+        [SuppressMessage("ReSharper", "InconsistentNaming")]
+        private class ComparableScore
```
bdach, line 168: The build warnings about "field never assigned to" are a bit annoying here. Easiest way to fix would probably be

```suggestion
        public class ComparableScore
```

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/VerifyImportedScoresCommand.cs
```diff
@@ -0,0 +1,180 @@
...
+using osu.Server.Queues.ScoreStatisticsProcessor.Commands.Queue;
+using osu.Server.Queues.ScoreStatisticsProcessor.Helpers;
+
+namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Maintenance
+{
+    [Command("verify-imported-scores", Description = "Verifies data and pp for already imported scores")]
+    public class VerifyImportedScoresCommand
+    {
+        /// <summary>
+        /// The high score ID to start deleting imported high scores from.
```
bdach, line 23:
```suggestion
        /// The high score ID to start verifying imported high scores from.
```

i sure hope the current version is incorrect :sweat_smile: :sweat_smile: :sweat_smile:

bdach:
Is this ready for review, or is there a possibility of further changes still?

peppy:
It's fine for now. The `Watch` command less so, but this one is holding up.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/VerifyImportedScoresCommand.cs
```diff
@@ -81,7 +81,12 @@ public async Task<int> OnExecuteAsync(CancellationToken cancellationToken)
 
                     if (importedScore.HighScore == null)
                     {
-                        // TODO: delete imported score
+                        await conn.ExecuteAsync("DELETE FROM scores WHERE id = @newId; DELETE FROM score_performance WHERE score_id = @newId; DELETE FROM solo_scores_legacy_id_map WHERE old_score_id = @oldId", new
```
bdach, line 84: ummmmmm wasn't `solo_scores_legacy_id_map` the wrong one and now it's `scores_legacy_id_map`...?

peppy: yep 😵‍💫

bdach: was *proooobably* lifted from

https://github.com/ppy/osu-queue-score-statistics/blob/e4e243313f8d7bf09e31453a6952aabc0f9f241d/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/BatchInserter.cs#L108

and we probably wanna fix that one too x_x

peppy: Yeah well fuck

peppy: Also it's missing a mode specification. Which will completely obliterate wrong scores in the worst case scenario.

bdach: I'm not seeing mode specs in d9cf7611f2f15639a657345ae9cbf82421cc7105 or b6ac0c900e49a7248323167b3e44e6b1c58786e6, is that something yet to be added still?

peppy: I'll push that now. Was testing on another branch.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/VerifyImportedScoresCommand.cs
```diff
@@ -0,0 +1,201 @@
...
+
+                        await conn.ExecuteAsync($"REPLACE INTO score_legacy_id_map (ruleset_id, old_score_id, score_id) VALUES ({importedScore.ruleset_id}, {importedScore.legacy_score_id}, {importedScore.id})");
+                    }
+
+                    if (importedScore.pp == null && importedScore.HighScore.pp != null)
+                    {
+                        Console.WriteLine($"{importedScore.id}: Performance entry missing!!");
+                        Interlocked.Increment(ref fail);
+
+                        await conn.ExecuteAsync($"REPLACE INTO score_performance VALUES ({importedScore.id}, {importedScore.HighScore.pp})");
```
bdach, line 114: The lack of an explicit column name list is a *bit* scary here but from testing on my local env it does seem like this should work as expected in this order...

bdach (approved):
i *think* this should work okay (not that that means much at this point anymore frankly)

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/VerifyImportedScoresCommand.cs
peppy: I had this marked for cleanup requirements, but the table doesn't exist anymore so that is no longer required.

PR #195, opened 2024-01-02
Implement all non-Hush-Hush medals

substanc3-dev:
This PR implements all non-secret medals:
 - Mod Introduction medals that happen when the player submits a score using a mod they haven't used alone before
 - Medals awarded for a pass/FC at a certain star rating
   - It may be desirable to check whether we want to trigger the 10\*std/8\*others medal when the player achieves a score of a higher SR (this is not how the lower-rated medals work, but might still be worth thinking about and/or checking osu-web-10 (medals in question would be `245` for std 10* and `102` as an example of the top-rated medal in an alternative ruleset))
 - Medals based around user statistics such as play count/hit count medals (this processor also handles stds combo-based medals)
 - Additional 10 beatmap pack medals that were missed in the initial implementation

This bring the total implemented medals to 205 out of 323, rest of which are Hush-Hush which probably needs some more consideration about how the code will be structured as I'm not sure it would make sense for that to be publicly available.


smoogipoo:
Nice. Was this referenced from somewhere?

substanc3-dev:
> Nice. Was this referenced from somewhere?

Unfortunately I don't think I have access to any code responsible for this in production (this is in osu-web-10), I mostly got this done by trial&error.
The medal IDs I grabbed from `data-initial-data` returned by osu-web in production, as I couldn't find a nicer place to get them, then generated a simple sql query so I could load these in a local full osu-web deploy (https://gist.github.com/substanc3-dev/ad0427cca98f85384e8bd37b9bef6c5b).

Mod Introduction medals I looked at random user profiles to see someone with a play with a mod but no mod medal, also cross-referenced with community resources about these, was pretty straight-forward.

SR medals were fairly simple, I found a user who got a higher-level medal without a lower one to confirm that's how that works (https://osu.ppy.sh/users/34729989/osu) then I tested getting an FC without having the pass medal first to see if they can be awarded at the same time (yes - https://osu.ppy.sh/users/6356412/taiko). The main thing I couldn't test is whether a 11 star+ play gives the 10 star medal (and whether a 9+ star play on taiko/etc. gives the 8 star one), because obviously, so I just implemented those the same way the lower-level medals work.

The other medals just trigger based on userstats and that was fairly straight-forward.

I've tested these both just getting then with lazer on my local osu-web deploy and I also of course wrote tests for these.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/MedalAwarders/ModIntroductionMedalAwarder.cs
```diff
@@ -0,0 +1,69 @@
...
+    [UsedImplicitly]
+    public class ModIntroductionMedalAwarder : IMedalAwarder
+    {
+        public bool RunOnFailedScores => false;
+
+        public IEnumerable<Medal> Check(SoloScoreInfo score, UserStats userStats, IEnumerable<Medal> medals, MySqlConnection conn, MySqlTransaction transaction)
+        {
+            // Ensure the mod is the only one selected
+            if (score.Mods.Length != 1)
+                yield break;
```
bdach, line 22: This is the legacy web implementation yes but I am not sure that it should continue to work this way. The primary counterexample is classic mod: I'm not sure that a classic + perfect play should not award the requisite medal.

I'm not sure there are any other exceptions one might want to allow other than classic but food for thought.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/MedalAwarders/ModIntroductionMedalAwarder.cs
```diff
@@ -0,0 +1,69 @@
...
+            foreach (var medal in medals)
+            {
+                if (checkMedal(score, medal, mod))
+                    yield return medal;
+            }
+        }
+
+        private bool checkMedal(SoloScoreInfo score, Medal medal, Mod mod)
+        {
+            return medal.achievement_id switch
```
bdach, line 40: This switchexpr usage is most definitely not required. I'm fine with it personally but others may not be.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/MedalAwarders/ModIntroductionMedalAwarder.cs
```diff
@@ -0,0 +1,69 @@
...
+                // Flashlight (Are You Afraid Of The Dark?)
+                125 => mod is ModFlashlight,
+                // Easy (Dial It Right Back)
+                126 => mod is ModEasy,
+                // No Fail (Risk Averse)
+                127 => mod is ModNoFail,
+                // Half Time (Slowboat)
+                128 => mod is ModHalfTime,
+                // Spun Out (Burned Out)
+                131 => mod is OsuModSpunOut,
```
bdach, line 63: Notably in this gap there are 2 more "mod introduction" medals that are marked inoperable on old web (129 => Relax, 130 => Autopilot). Unsure if we want to add these.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/MedalAwarders/StarRatingMedalAwarder.cs
```diff
@@ -0,0 +1,154 @@
...
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace osu.Server.Queues.ScoreStatisticsProcessor.Processors.MedalAwarders
+{
+    [UsedImplicitly]
+    [SuppressMessage("ReSharper", "MergeIntoPattern")]
```
bdach, line 20:
```suggestion
```
does nothing

osu.Server.Queues.ScoreStatisticsProcessor/Processors/MedalAwarders/StarRatingMedalAwarder.cs
```diff
@@ -0,0 +1,154 @@
...
+        private async Task<IEnumerable<Medal>> checkAsync(SoloScoreInfo score, IEnumerable<Medal> medals, MySqlConnection conn, MySqlTransaction transaction)
+        {
+            List<Medal> earnedMedals = new List<Medal>();
+
+            Ruleset ruleset = LegacyRulesetHelper.GetRulesetFromLegacyId(score.RulesetID);
+            Mod[] mods = score.Mods.Select(m => m.ToMod(ruleset)).ToArray();
+
+            // Ensure the score isn't using any difficulty reducing mods
+            if (mods.Any(x => x is ModEasy || x is ModNoFail || x is ModHalfTime))
+                return earnedMedals;
```
bdach, line 41: Spun Out too.

I'd recommend splitting this to a helper method because this is used very frequently in medal awarding code.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/MedalAwarders/StarRatingMedalAwarder.cs
```diff
@@ -0,0 +1,154 @@
...
+
+            try
+            {
+                beatmapStore ??= await BeatmapStore.CreateAsync(conn, transaction);
+
+                Beatmap? beatmap = await beatmapStore.GetBeatmapAsync(score.BeatmapID, conn, transaction);
+                if (beatmap == null)
+                    return earnedMedals;
```
bdach, line 49: Notably the old implementation excludes qualified and loved beatmaps here:

```php
if ($diffReduction || $beatmap['approved'] > 2) break;
```

Unsure whether that's something to preserve.

This applies to all of the medal ranges in this file.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/MedalAwarders/StarRatingMedalAwarder.cs
```diff
@@ -0,0 +1,154 @@
...
+        // Checks for medals in a 1-8* range, which tend to be sequential IDs
+        private bool checkMedalRange(int medalIdStart, int medalId, double starRating)
+        {
+            if (medalId < medalIdStart || medalId > medalIdStart + 7)
+                return false;
+
+            return checkSR(starRating, medalId - medalIdStart + 1);
+        }
+
+        private bool checkSR(double starRating, int expected)
```
bdach, line 151: Please rename to `checkStarRating()`

osu.Server.Queues.ScoreStatisticsProcessor/Processors/MedalAwarders/StarRatingMedalAwarder.cs
```diff
@@ -0,0 +1,154 @@
...
+
+        private bool checkMedalPass(SoloScoreInfo score, Medal medal, double starRating)
+        {
+            // osu!standard 1-8*
+            if (checkMedalRange(55, medal.achievement_id, starRating))
+                return true;
+
+            // osu!taiko 1-8*
+            if (checkMedalRange(71, medal.achievement_id, starRating))
+                return true;
```
bdach, line 90: https://osu.ppy.sh/beatmapsets/2626#taiko/19990 is explicitly disallowed for this medal range.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/MedalAwarders/StarRatingMedalAwarder.cs
```diff
@@ -0,0 +1,154 @@
...
+
+            // osu!catch 1-8*
+            if (checkMedalRange(79, medal.achievement_id, starRating))
+                return true;
+
+            // osu!mania 1-8*
+            if (checkMedalRange(87, medal.achievement_id, starRating))
+                return true;
```
bdach, line 98: Key mods and "dual stages" are disallowed for this medal range.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/MedalAwarders/StarRatingMedalAwarder.cs
```diff
@@ -0,0 +1,154 @@
...
+
+        private bool checkMedalFc(SoloScoreInfo score, Medal medal, double starRating)
+        {
+            // osu!standard 1-8*
+            if (checkMedalRange(63, medal.achievement_id, starRating))
+                return true;
+
+            // osu!taiko 1-8*
+            if (checkMedalRange(95, medal.achievement_id, starRating))
+                return true;
```
bdach, line 120: https://osu.ppy.sh/beatmapsets/2626#taiko/19990 is explicitly disallowed for this medal range.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/MedalAwarders/StarRatingMedalAwarder.cs
```diff
@@ -0,0 +1,154 @@
...
+
+            // osu!catch 1-8*
+            if (checkMedalRange(103, medal.achievement_id, starRating))
+                return true;
+
+            // osu!mania 1-8*
+            if (checkMedalRange(111, medal.achievement_id, starRating))
+                return true;
```
bdach, line 128: Key mods and "dual stages" are disallowed for this medal range.

bdach (changes requested):
Initial cross-reference pass

bdach:
Remaining pack IDs:

| medal ID | pack ID |
| -: | :- |
| 308 | 2709 |
| 309 | 2710 |
| 310 | 2711 |
| 311 | 2712 |
| 312 | 2716 |
| 313 | 2720 |
| 314 | 2724 |
| 315 | 2738 |
| 316 | 2732 |
| 325 | 3122 OR 3123 OR 3124 OR 3125[^1] |

[^1]: requires any of the 4 packs cleared

osu.Server.Queues.ScoreStatisticsProcessor/Processors/MedalAwarders/ModIntroductionMedalAwarder.cs
peppy: At least a TODO comment mentioning this. I think we'll want to create these medals for lazer.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/MedalAwarders/ModIntroductionMedalAwarder.cs
substanc3-dev: Classic mod is currently disallowed for ranking (other than for legacy score), which is how I implemented this.

I tried to do this in a way where this would be straight-forward to expand with other mods when/if desired.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/MedalAwarders/ModIntroductionMedalAwarder.cs
substanc3-dev: While it's not required for the existing medals per se, I would argue it's pretty nice to have regardless, considering mods can now have customized settings, so the conditions here could get more complex (for example a 2x dt medal, or special DA medals etc.)

osu.Server.Queues.ScoreStatisticsProcessor/Processors/MedalAwarders/ModIntroductionMedalAwarder.cs
substanc3-dev: Added a comment for this.

substanc3-dev:
Thanks for the cross-reference and the beatmap pack information. I've implemented the rest of the beatmap pack medals and tried to address points brought up in review + bring medal conditions to parity with stable.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/MedalAwarders/StarRatingMedalAwarder.cs
substanc3-dev: Addressed as such:

https://github.com/ppy/osu-queue-score-statistics/blob/d5f1eb7d096c083510a75951d52a0afec8858f4a/osu.Server.Queues.ScoreStatisticsProcessor/Processors/MedalAwarders/StarRatingMedalAwarder.cs#L44-L46

osu.Server.Queues.ScoreStatisticsProcessor/Processors/MedalAwarders/StarRatingMedalAwarder.cs
substanc3-dev: Added this, as especially Loved maps tend to sometimes have weird elements in them that break SR, so this is likely desired.

https://github.com/ppy/osu-queue-score-statistics/blob/d5f1eb7d096c083510a75951d52a0afec8858f4a/osu.Server.Queues.ScoreStatisticsProcessor/Processors/MedalAwarders/StarRatingMedalAwarder.cs#L56-L58

osu.Server.Queues.ScoreStatisticsProcessor/Processors/MedalAwarders/ModIntroductionMedalAwarder.cs
bdach: Ranking is not medal granting though. The two are independent. Classic mod plays will submit and show on leaderboards but will not give pp.

substanc3-dev: Okay, changed this to also apply on non-legacy classic scores.

osu.Server.Queues.ScoreStatisticsProcessor/Helpers/ModHelper.cs
```diff
@@ -0,0 +1,28 @@
...
+                case ModEasy:
+                case ModNoFail:
+                case ModHalfTime:
+                case OsuModSpunOut:
+                    return true;
+
+                // Difficulty adjust can be used to decrease the difficulty of the map,
+                // so let's consider it difficulty reducing for now.
+                case ModDifficultyAdjust difficultyAdjust:
+                    return !difficultyAdjust.UsesDefaultConfiguration;
```
bdach, line 21: If we're going there we probably need to exclude a lot more...

- relax
- autopilot
- daycore

etc. etc.

Tempted to just say we should be switching this to something like `mod.Type == ModType.DifficultyReduction || mod.Type == ModType.Automation` at that point. Maybe `ModType.Fun` too, not sure.

bdach (commented):
Test coverage of all of the edge cases introduced after cross-check would be appreciated

osu.Server.Queues.ScoreStatisticsProcessor/Helpers/ModHelper.cs
substanc3-dev: I made it check for `ModType.DifficultyReduction` and `ModType.Automation`, most of `ModType.Fun` except a couple, then also exclude some other mods (DA and TP). I think this should cover it pretty well.

bdach: I'm still not sure this goes far enough. Most conversion mods (random, invert, mirror...) should probably be disallowed too.

In fact, one point to this is that because this is looking up difficulty attributes from database (until we get realtime difficulty working), _no_ new mods should be allowed for star rating medals (unless one makes an executive decision that they don't affect anything), because we _don't have_ accurate star ratings with arbitrary mods. Only the combinations that stable allows should be supported.

So for now we probably want to swap this blocklist to an allowlist instead that matches what stable permits exactly.

osu.Server.Queues.ScoreStatisticsProcessor.Tests/MedalProcessorTests.cs
```diff
@@ -238,6 +243,376 @@ public void TestNoReductionModsPack()
...
+
+            SetScoreForBeatmap(beatmap.beatmap_id, s => s.Score.ScoreInfo.Mods = new[] { new APIMod(new OsuModDoubleTime()), new APIMod(new OsuModClassic()) });
+
+            // When the beatmap is completed with double time and classic, the medal should be awarded.
+            assertAwarded(medal_id);
+        }
+
+        /// <summary>
+        /// This tests the star rating medals, both pass and FC.
+        /// It also tests the fact that these medals should not be awarded, in case there are other mods enabled.
```
bdach, line 293: copy-paste error?

osu.Server.Queues.ScoreStatisticsProcessor.Tests/MedalProcessorTests.cs
```diff
@@ -238,6 +243,376 @@ public void TestNoReductionModsPack()
...
+        /// </summary>
+        [Fact]
+        public void TestStarRatingMedals()
+        {
+            const int medal_id_pass = 59;
+            const int medal_id_fc = 67;
+
+            // Beatmap ID 2 to ensure we don't use cached star rating info
```
bdach, line 301: What does this mean?

osu.Server.Queues.ScoreStatisticsProcessor.Tests/MedalProcessorTests.cs
```diff
@@ -238,6 +243,376 @@ public void TestNoReductionModsPack()
...
+            assertNoneAwarded();
+
+            // Set a score without restricted mods, which should award the medal
+            SetScoreForBeatmap(beatmap.beatmap_id, s => s.Score.ruleset_id = 3);
+
+            assertAwarded(medal_id_5_star);
+        }
+
+        /// <summary>
+        /// This tests the star rating medals, to make sure a loved maps don't trigger medals.
```
bdach, line 475:
```suggestion
        /// This tests the star rating medals, to make sure loved maps don't trigger medals.

```

osu.Server.Queues.ScoreStatisticsProcessor.Tests/MedalProcessorTests.cs
```diff
@@ -238,6 +243,376 @@ public void TestNoReductionModsPack()
...
+
+            assertNoneAwarded();
+            SetScoreForBeatmap(beatmap.beatmap_id, s => s.Score.ScoreInfo.MaxCombo = 499);
+
+            // After passing the beatmap without getting 500 combo, the medal shouldn't be awarded.
+            assertNoneAwarded();
+
+            SetScoreForBeatmap(beatmap.beatmap_id, s => s.Score.ScoreInfo.MaxCombo = 500);
+
+            // Once the beatmap is FC'd with enough combo, the medal may be awarded.
```
bdach, line 520: What is this mention of "FC"? The medal does not check that?

bdach (changes requested):

osu.Server.Queues.ScoreStatisticsProcessor.Tests/MedalProcessorTests.cs
substanc3-dev: Hopefully made more clear by the new comment

osu.Server.Queues.ScoreStatisticsProcessor/Helpers/ModHelper.cs
substanc3-dev: Implemented as an allow-list for now, can be adjusted whenever SR has support for the other mods.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/MedalAwarders/StarRatingMedalAwarder.cs
```diff
@@ -0,0 +1,179 @@
...
+            {
+                beatmapStore ??= await BeatmapStore.CreateAsync(conn, transaction);
+
+                Beatmap? beatmap = await beatmapStore.GetBeatmapAsync(score.BeatmapID, conn, transaction);
+                if (beatmap == null)
+                    return Enumerable.Empty<Medal>();
+
+                // Make sure the map isn't Qualified or Loved, as those maps may occasionally have SR-breaking/aspire aspects
+                if (beatmap.approved == BeatmapOnlineStatus.Qualified || beatmap.approved == BeatmapOnlineStatus.Loved)
+                    return Enumerable.Empty<Medal>();
```
Walavouchey, line 58: i'm not sure i get this logic. aren't graveyarded maps also prone to sr-breaking aspects?

bdach: Anything that isn't ranked should not be giving any medals period. Likely a general oversight at a higher level.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/MedalAwarders/ModIntroductionMedalAwarder.cs
cl8n: imo the correct mods to exclude here would be Classic as well as all "System" type mods. the point of these medals, how i understand them, is that the player tried the mod by itself to see how it works -- but the player has no direct control over system mods.

bdach: @cl8n should be addressed with 2a8da5deaca77490c653af214a0f87e3f20db25f (with test coverage at 23774c1651c873ef249577459aeb8c445126f092

osu.Server.Queues.ScoreStatisticsProcessor/Helpers/ModHelper.cs
bdach: I've reformulated this slightly with the new `Ranked` property in d25b6fddc48e8f9e654fac912064e76dcfe51d05 (test coverage at 532ec32ad6baee2cf88173283828173b53c516f0). Not sure how this is going to hold up long-term, but let's start with this and see where this goes.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/MedalAwarders/StarRatingMedalAwarder.cs
bdach: @Walavouchey addressed in 85137b7affa0251115d0e907db5b3cae84616e39 (with test coverage in f22be2b2b9df6c81e8a83758bc79eef309864f32).

There is _one_ medal that could at a time be granted on unranked beatmaps. That was this December, and it's now turned off to the best of my knowledge. We don't talk about this one, _wink wink_. (And even if we were to, it wouldn't be implemented in this public reposiory.)

bdach:
I've done a merge conflict + review resolution pass on this in hopes of moving this forward. I'm way too spent to give it a proper once-over right now, so I'll do that tomorrow if I don't get overtaken.

Reviews from everyone involved here up until now welcome.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/MedalAwarders/ModIntroductionMedalAwarder.cs
bdach: Let's keep this for now.

tsunyoku (approved):
I've been watching quietly thus far but looks good.

bdach:
I've given this another read today. Have applied some last-minute changes (including a medal that just dropped today), but other than that I'm pretty happy where this is right now.

Doing a sweep over all medals, there are actually five medals that are not implemented right now even with this change and are not hush-hush, and those would be the global rank related medals (ids 50-54). I think we can do these as a follow-up, and also maybe we should do those _after_ the whole shebang with people jumping and dropping in ranks depending on which client they used last is sorted.

Leaving an approval, but given the extent of changes and some liberties taken when it comes to some new decisions as to whether award a medal or not that were not considerations in stable, I would like a second review from either @peppy or @smoogipoo as a confirmation.

bdach (approved):

substanc3-dev:
> Doing a sweep over all medals, there are actually five medals that are not implemented right now even with this change and are not hush-hush, and those would be the global rank related medals (ids 50-54). I think we can do these as a follow-up, and also maybe we should do those _after_ the whole shebang with people jumping and dropping in ranks depending on which client they used last is sorted.

Good catch, not sure how I've missed it tbh, I vaguely remember looking into that. I will look into creating that change after this base PR is merged, should just be a matter of adding an ordering dependency for `UserTotalPerformanceProcessor` into `MedalProcessor` and then adding the respective medals into `StatisticsMedalAwarder`.

peppy:
One thing about this is, while we can make it live, there's still no handling in-game of incoming achievement events. I do believe we can already get these from the notification server flow (ie. osu-web is already sending *some* details throuhg).

bdach:
> there's still no handling in-game of incoming achievement events

Well aware. That's what I was going to work on at some point this or next week.

Also I believe even the medals that are already implemented don't seem to work from last I tested. Not sure what that's about, I'll re-check at some point, but it may be something in the legacy IO innards that I'm not able to debug.

substanc3-dev:
> > there's still no handling in-game of incoming achievement events
> 
> Well aware. That's what I was going to work on at some point this or next week.
> 
> Also I believe even the medals that are already implemented don't seem to work from last time things were tested. Not sure what that's about, I'll re-check at some point, but it may be something in the legacy IO innards that I'm not able to debug.

When I tested this at the time of making the PR it worked fine, the medals got unlocked on the website and the notifications were delivered (on the web). I can try re-testing this again with latest osu-web as a lot has changed, but afaik this mostly came down to properly setting up LIO endpoint in `LegacyDatabaseHelper` + actually setting up the `SHARED_INTEROP_SECRET` for the instance.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/MedalProcessor.cs
```diff
@@ -43,12 +43,19 @@ static MedalProcessor()
 
         public bool RunOnLegacyScores => false;
 
+        // ReSharper disable once UselessBinaryOperation
+        // This processor needs to run after the play count and hit statistics have been applied.
+        public int Order => Math.Max(PlayCountProcessor.ORDER, HitStatisticsProcessor.ORDER) + 1;
```
peppy, line 48: Would it not make more sense to just make this always run last? ie. give it a very very high `ORDER`?

peppy: Something like

```diff
diff --git a/osu.Server.Queues.ScoreStatisticsProcessor/Processors/MedalProcessor.cs b/osu.Server.Queues.ScoreStatisticsProcessor/Processors/MedalProcessor.cs
index 22afa78..c93862e 100644
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Processors/MedalProcessor.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Processors/MedalProcessor.cs
@@ -45,7 +45,7 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Processors
 
         // ReSharper disable once UselessBinaryOperation
         // This processor needs to run after the play count and hit statistics have been applied.
-        public int Order => Math.Max(PlayCountProcessor.ORDER, HitStatisticsProcessor.ORDER) + 1;
+        public int Order => int.MaxValue;
 
         public void RevertFromUserStats(SoloScoreInfo score, UserStats userStats, int previousVersion, MySqlConnection conn, MySqlTransaction transaction)
         {

```

bdach: Wouldn't hurt I suppose.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/MedalAwarders/StarRatingMedalAwarder.cs
```diff
@@ -0,0 +1,179 @@
...
+            if (mods.Any(MedalHelpers.IsDifficultyReductionMod))
+                return Enumerable.Empty<Medal>();
+
+            // On mania, these medals cannot be triggered with key mods and Dual Stages
+            if (score.RulesetID == 3 && mods.Any(isManiaDisallowedMod))
+                return Enumerable.Empty<Medal>();
+
+            try
+            {
+                beatmapStore ??= await BeatmapStore.CreateAsync(conn, transaction);
```
peppy, line 50: Not sure about creating a second store here, rather than sharing the existing one (see https://github.com/ppy/osu-queue-score-statistics/blob/9f99f91f31087ff70d060d27ebd120f9b78f2195/osu.Server.Queues.ScoreStatisticsProcessor/Processors/ScorePerformanceProcessor.cs#L118). I'll probably look to combine these usages.

bdach: Funny you should mention that, I did that in https://github.com/ppy/osu-queue-score-statistics/pull/230, so for sure.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/MedalAwarders/StatisticsMedalAwarder.cs
```diff
@@ -0,0 +1,80 @@
...
+                    yield return medal;
+            }
+        }
+
+        private bool checkMedal(Medal medal, MedalAwarderContext context)
+        {
+            var score = context.Score;
+            var stats = context.UserStats;
+
+            return medal.achievement_id switch
```
peppy, line 26: I'll let these switch statements pass for now, but reserve the right to convert them away from expression types in the future. Still not sold and don't think it helps that much with legibility.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/MedalAwarders/StarRatingMedalAwarder.cs
```diff
@@ -0,0 +1,173 @@
...
+                    foreach (var medal in medals)
+                    {
+                        if (checkMedalFc(context.Score, medal, difficultyAttributes.StarRating))
+                            earnedMedals.Add(medal);
+                    }
+                }
+            }
+            catch (Exception ex)
+            {
+                await Console.Error.WriteLineAsync($"{context.Score.ID} failed in StarRatingMedalAwarder with: {ex}");
```
peppy, line 80: Do we really want this? It stops the queue item from throwing itself, triggering retries and potentially alerting from queue processor restarts.

I bring this up because it's also stopping us from using `yield return`, which simplifies code considerably.

@bdach

bdach: Difficult to anticipate what the failure modes here are going to be but you're right, probably better to be louder about it.

peppy (approved):

PR #196, opened 2024-01-03
Fix some entries not waiting for performance population before importing

peppy:
Simplest fix for now. Has been deployed a couple of days and is running well.


bdach (approved):

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/WatchHighScoresCommand.cs
```diff
@@ -132,6 +132,15 @@ public async Task<int> OnExecuteAsync(CancellationToken cancellationToken)
                                              .OrderBy(s => s.queue_id)
                                              .ToArray();
 
+                    var pendingProcessing = highScores.FirstOrDefault(s => s.status == 0);
```
bdach, line 135: I was side-eyeing this because `status` is declared as `byte?` so wasn't sure whether it was possible to have it `null`. But it's not, because the query is a `SELECT * FROM osu.score_process_queue`, so if a row is returned here, it _must_ have the `score_process_queue` members populated

bdach (approved):
no never mind all good

PR #201, opened 2024-01-17
Fix some issues with watch mode

peppy:
Now that I'm running it... I found some issues 😅 .

Confirmed working well via deploy-to-production testing technique.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/WatchHighScoresCommand.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/WatchHighScoresCommand.cs
@@ -125,7 +125,7 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Queue
                     HighScore[] highScores = (await dbMainQuery.QueryAsync<HighScore>(
                                                  "SELECT q.*, h.*, s.id as new_id FROM osu.score_process_queue q "
                                                  + $"LEFT JOIN {highScoreTable} h USING (score_id) "
-                                                 + $"LEFT JOIN scores s ON h.score_id = s.legacy_score_id AND s.ruleset_id = {RulesetId} "
+                                                 + $"LEFT JOIN scores s ON q.score_id = s.legacy_score_id AND s.ruleset_id = {RulesetId} "
                                                  + $"WHERE queue_id >= @lastQueueId AND mode = {RulesetId} ORDER BY queue_id LIMIT 50", new
                                                  {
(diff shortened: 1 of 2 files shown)


osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/WatchHighScoresCommand.cs
```diff
@@ -123,9 +123,9 @@ public async Task<int> OnExecuteAsync(CancellationToken cancellationToken)
                 while (!cancellationToken.IsCancellationRequested)
                 {
                     HighScore[] highScores = (await dbMainQuery.QueryAsync<HighScore>(
-                                                 "SELECT h.*, s.id as new_id FROM osu.score_process_queue "
+                                                 "SELECT q.*, h.*, s.id as new_id FROM osu.score_process_queue q "
```
peppy, line 126: This is a bit of cheating: we want to get `score_id` from the queue table (because it will be null in a deletion from the `highScoreTable`). Dapper will use the first available, which becomes the id from the queue.

bdach: that makes me uncomfortable but not deeply enough to make a big deal out of it. let's see if i'm wrong about that

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/WatchHighScoresCommand.cs
```diff
@@ -123,9 +123,9 @@ public async Task<int> OnExecuteAsync(CancellationToken cancellationToken)
...
                                                  + $"LEFT JOIN {highScoreTable} h USING (score_id) "
-                                                 + $"LEFT JOIN scores s ON h.score_id = s.legacy_score_id AND s.ruleset_id = {RulesetId} "
+                                                 + $"LEFT JOIN scores s ON q.score_id = s.legacy_score_id AND s.ruleset_id = {RulesetId} "
```
bdach, line 128: is this join change meaningful? would like to think that it's not...

peppy: It actually is, because on deletions the `h` row will be null. Mentioned in https://github.com/ppy/osu-queue-score-statistics/pull/201/commits/f5556a09cd30038eaee89c4357ffa86dfecbda03 commit message.

bdach (approved):

PR #202, opened 2024-01-19
Update verify scores command to handle total score and legacy total score

peppy:
Testing in try run on production. Performance is limited to single thread but can easily be improved by threading.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/VerifyImportedScoresCommand.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/VerifyImportedScoresCommand.cs
@@ -10,6 +10,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Dapper;
 using McMaster.Extensions.CommandLineUtils;
+using osu.Game.Scoring;
 using osu.Server.QueueProcessor;
 using osu.Server.Queues.ScoreStatisticsProcessor.Commands.Queue;
 using osu.Server.Queues.ScoreStatisticsProcessor.Helpers;
@@ -80,6 +81,18 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Maintenance
 
                 elasticItems.Clear();
 
+                Parallel.ForEach(importedScores, new ParallelOptions
+                {
+                    MaxDegreeOfParallelism = Environment.ProcessorCount,
+                }, importedScore =>
+                {
+                    if (importedScore.legacy_score_id == null) return;
+
+                    if (importedScore.HighScore == null) return;
+
+                    importedScore.ReferenceScore = BatchInserter.CreateReferenceScore(importedScore.ruleset_id, importedScore.HighScore);
+                });
+
                 foreach (var importedScore in importedScores)
                 {
                     if (importedScore.legacy_score_id == null) continue;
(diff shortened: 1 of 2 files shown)


osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/BatchInserter.cs
```diff
@@ -379,6 +373,36 @@ private static LegacyBeatmapConversionDifficultyInfo getDificultyInfo(int beatma
...
+                maxBasicResult = ruleset.GetHitResults().Where(h => h.result.IsBasic()).Select(h => h.result).MaxBy(scoreProcessor.GetBaseScoreForResult);
+
+            cache = new RulesetCache(ruleset, maxBasicResult, ruleset.CreateMod<ModClassic>()!);
+
+            ruleset_cache[rulesetId] = cache;
+
+            return cache;
+        }
+
+        private record RulesetCache(Ruleset Ruleset, HitResult MaxBasicResult, Mod ClassicMod);
```
bdach, line 396: `record`? in _your_ code? well this is an unexpected development :smile:

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/VerifyImportedScoresCommand.cs
```diff
@@ -39,11 +49,14 @@ public async Task<int> OnExecuteAsync(CancellationToken cancellationToken)
...
 
+            if (DryRun)
+                Console.WriteLine("RUNNING IN DRY RUN MODE.");
+
             using var conn = DatabaseAccess.GetConnection();
 
             while (!cancellationToken.IsCancellationRequested)
             {
-                List<ElasticQueuePusher.ElasticScoreItem> elasticItems = new List<ElasticQueuePusher.ElasticScoreItem>();
+                HashSet<ElasticQueuePusher.ElasticScoreItem> elasticItems = new HashSet<ElasticQueuePusher.ElasticScoreItem>();
```
bdach, line 59: Making this a `HashSet<>` seems dubious given that `ElasticScoreItem` is a plain class with no equality members overridden. So if the intent here was to deduplicate anything, then I'm pretty sure that will not happen, because the default comparison method for a plain class without equality members overridden is *reference* equality.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/BatchInserter.cs
peppy: I was just matching suit below. 🤷

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/VerifyImportedScoresCommand.cs
peppy: Yeah it kinda was for deduping, but it's not a huge deal for now.

I'll change the logic to dedupe because it's simple to do and probably reads better.

peppy: In fact, I think there was a missing index op for the new cases I added, so I'll need to run manual reindexing.

bdach: > I'll change the logic to dedupe

I got a re-review request, but nothing seems to have happened in this respect (yet?)

peppy: https://github.com/peppy/osu-queue-score-statistics/commit/5ba0eda4ef638ac9841f39bb9e0ef3b7459094c9 is intended to "dedupe", aka move the index request to the `finally` block.

bdach (approved):

PR #203, opened 2024-01-22
Use updated accuracy & grade definitions when importing stable scores

bdach:
- Will close https://github.com/ppy/osu/issues/26595
- [x] Depends on https://github.com/ppy/osu/pull/26630
- [x] Depends on game package w/ above

~~**Caution: This is very untested.** It only works in my head. I would test but right now my box is working as a heater doing diffcalc spreadsheets, and I'd rather not push my luck and have it hang on me. I'm PRing early for visibility and for visibility alone. And testing convenience I guess since I will want to run score sheets using this diff to verify that I didn't break anything.~~

As requested in https://github.com/ppy/osu/issues/26595#issuecomment-1903767921, this includes a rank check in the verify command from https://github.com/ppy/osu-queue-score-statistics/commit/68e53cda1583d44ec0466ba23a69348eb93edeba. But with fixes around quoting the `rank` column name (and actually selecting it from DB).

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor.Tests/osu.Server.Queues.ScoreStatisticsProcessor.Tests.csproj
+++ b/osu.Server.Queues.ScoreStatisticsProcessor.Tests/osu.Server.Queues.ScoreStatisticsProcessor.Tests.csproj
@@ -9,8 +9,8 @@
     <ItemGroup>
         <PackageReference Include="DeepEqual" Version="4.2.1" />
         <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.8.0" />
-        <PackageReference Include="xunit" Version="2.6.3" />
-        <PackageReference Include="xunit.runner.visualstudio" Version="2.5.5">
+        <PackageReference Include="xunit" Version="2.6.6" />
+        <PackageReference Include="xunit.runner.visualstudio" Version="2.5.6">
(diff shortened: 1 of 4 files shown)


bdach:
I have ran the verify command now and it seems to be doing what it is supposed to be doing.

I hope to achieve the best approximation of full verification that this changes nothing wrt score conversions via the score spreadsheets I intend to run in the game-side change (https://github.com/ppy/osu/pull/26630#issuecomment-1905939974), but for whatever it's worth this should now be at least reviewable.

peppy:
@bdach for my own sanity, is this not going to require rewriting every database row to update accuracies? and if so, we need to add this to the verify command yeah?

bdach:
> is this not going to require rewriting every database row to update accuracies?

I don't believe so, because accuracy was being recomputed before anyway.

This call on master:

https://github.com/ppy/osu-queue-score-statistics/blob/5c23e0f02b67c6089922ea553ce03894f2690011/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/BatchInserter.cs#L313

would also recalc accuracy (see [game-side source for the method at 2024.113.0](https://github.com/ppy/osu/blob/3d3f58c2523f519504d9156a684538e2aa0c094c/osu.Game/Database/StandardisedScoreMigrationTools.cs#L253-L254)).

peppy (approved):

PR #205, opened 2024-01-26
Optimise user PP total calculation

peppy:
Temporary optimisations being rolled out pending a refactor of caching mechanisms.

The beatmap join adds around 30% overhead to query, but is very simple so preferring for now.

Removed build checks because we'll do that using the `ranked` flag instead.

Doesn't need merge necessarily, just for visibility.


osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserTotalPerformanceProcessor.cs
```diff
@@ -68,96 +62,53 @@ public void ApplyGlobal(SoloScoreInfo score, MySqlConnection conn)
...
-                lastStoreRefresh = currentTimestamp;
-            }
 
             List<SoloScore> scores = (await connection.QueryAsync<SoloScore>(
-                "SELECT * FROM scores WHERE `user_id` = @UserId AND `ruleset_id` = @RulesetId AND `preserve` = 1", new
+                "SELECT beatmap_id, pp, accuracy FROM scores s "
+                + "JOIN osu_beatmaps USING (beatmap_id) WHERE "
+                // TODO: This adds a small amount of overhead to the query.
+                // Replace this with a cached lookup once we have a good system in place.
+                + "`approved` in (1,2) AND "
```
bdach, line 71: Is this still relevant? Weren't scores on other beatmaps going to have `ranked = 0` set?

Although as far as I recall we only talked about doing that for qualified, but there are also other cases like loved that were _not_ talked about explicitly...

osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserTotalPerformanceProcessor.cs
```diff
@@ -68,96 +62,53 @@ public void ApplyGlobal(SoloScoreInfo score, MySqlConnection conn)
...
-                if (s.legacy_score_id != null)
-                    return false;
-
-                // Some older lazer scores don't have build IDs.
-                if (s.build_id == null)
-                    return true;
-
-                // Performance needs to be allowed for the build.
-                if (buildStore.GetBuild(s.build_id.Value)?.allow_performance != true)
-                    return true;
```
bdach, line 127: I don't see any of these conditions reflected in the sql query, is it safe to assume `ranked` flag will cover that?

They don't seem to be necessarily very relevant right now (legacy scores are still processed by `osu-performance` and imported directly, old lazer scores don't exist in new tables) but yeah.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserTotalPerformanceProcessor.cs
peppy: it's not. i think this branch may be outdated me as i have been using my other combined branch

osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserTotalPerformanceProcessor.cs
peppy: yep

osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserTotalPerformanceProcessor.cs
peppy: I've pushed my off-branch changes to this branch for reference.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserTotalPerformanceProcessor.cs
peppy: This has been added to the score performance processor now.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserTotalPerformanceProcessor.cs
peppy: This is also set in the score performance processor now, and updates the `ranked` flag.

osu.Server.Queues.ScoreStatisticsProcessor/Models/SoloScore.cs
```diff
@@ -54,7 +54,7 @@ public string data
         public double? pp { get; set; }
 
         public ulong? legacy_score_id { get; set; }
-        public uint? legacy_total_score { get; set; }
+        public uint legacy_total_score { get; set; }
```
bdach, line 57: is this what we're going forward with just to fix tests? even though it is the osu-web image that isn't matching production? not even a todo in sight to get this fixed later?

peppy: apparently it was an oversight and it's non-null on production and osu-web. my local test setup was fucked.

bdach: > i edited my message showing actual production `CREATE TABLE` soon after posting if you want to re-check

well ok thanks for clarifying. saves me from wondering if i had a stroke earlier today

peppy:
https://github.com/ppy/osu-queue-score-statistics/pull/205/commits/8f51c672ae058b8146d3119d6a0b113193bec2b7 fixes the issue i mentioned in infra. piled on this PR because it's 99% related.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/ScorePerformanceProcessor.cs
```diff
@@ -108,33 +111,67 @@ public async Task ProcessScoreAsync(SoloScoreInfo score, MySqlConnection connect
             if (!score.Passed)
                 return;
 
-            beatmapStore ??= await BeatmapStore.CreateAsync(connection, transaction);
+            long currentTimestamp = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds();
+
+            if (buildStore == null || beatmapStore == null || currentTimestamp - lastStoreRefresh > 60)
```
bdach, line 116: Uh..... if this meant to refresh the store every minute, then that's not what this says is it? This refreshes the store every 60 _milliseconds_.

```suggestion
            if (buildStore == null || beatmapStore == null || currentTimestamp - lastStoreRefresh > 60_000)
```

is what you probably want if you intended the former.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/RecalculateTotalUserPerformanceCommand.cs
```diff
@@ -0,0 +1,46 @@
...
+                    var score2 = db.QuerySingle<SoloScore>($"SELECT id, user_id, ruleset_id FROM scores WHERE id = {score.id}");
+                    if (cancellationToken.IsCancellationRequested)
+                        break;
+
+                    sw.Restart();
+
+                    processor.ApplyToUserStats(score2.ToScoreInfo(), new UserStatsOsu { user_id = (int)score2.user_id }, db, null!);
+
+                    Console.WriteLine($"Recalculating {score2.user_id} took {sw.ElapsedMilliseconds:N0} ms");
+                }
```
bdach, line 39: Is this supposed to be here? I don't even know what this is attempting to do because it's making very little sense. Outer loop is over every non-imported score, inner query... fetches two more columns of the same row, and then something is happening with `ApplyToUserStats()` but the ruleset is... hardcoded by specifying `UserStatsOsu`? Very confused. I don't think this ever has an actual use for anything.

bdach (changes requested):

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/RecalculateTotalUserPerformanceCommand.cs
peppy: This was just used to benchmark the performance of the processor. I think it can be removed because as you say, it's likely useless.

bdach (approved):

PR #206, opened 2024-01-29
Add command to watch non-high-scores

peppy:
Until now, `osu_scores` table has been used to show "recent plays" on user profiles. This doesn't contain lazer scores.

To allow osu-web to read from only a single source of truth (the new `scores` table) we must import these scores from the old tables.

This watcher handles that process. It's written to be as simple as possible to get this out for the initial launch.

- This only handles non-high-scores. To make this work, there's a 5 second delay applied, to ensure that the `high_score_id` column has been populated by `osu-web-10` (it doesn't guard the insertion into `osu_scores` and `osu_scores_high` with a transaction, so this is best-effort).
- There's no gap / resume / whatever support. If this isn't running, scores are just not imported. These scores are throwaway (they get partition cycled and are only expected to be displayed for 24 hours), and the assumption is that these importers will be setup once and never fail. If that's not the case we can reconsider.
- `total_score` (ie. the converted one) may be 0 for some scores which don't have score attributes populated yet. We'll fix this later. Probably after users start noticing it.

Sorry for the fact that this makes `HighScore` even weirder. Is what it is I think.

This is already production tested and will be deployed ahead of this being merged.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/BatchInserter.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/BatchInserter.cs
@@ -75,7 +75,9 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Queue
 
             int rulesetId = ruleset.RulesetInfo.OnlineID;
 
-            StringBuilder insertBuilder = new StringBuilder("INSERT INTO scores (`user_id`, `ruleset_id`, `beatmap_id`, `has_replay`, `preserve`, `rank`, `passed`, `accuracy`, `max_combo`, `total_score`, `data`, `pp`, `legacy_score_id`, `legacy_total_score`, `ended_at`, `unix_updated_at`) VALUES ");
+            StringBuilder insertBuilder =
+                new StringBuilder(
+                    "INSERT INTO scores (`user_id`, `ruleset_id`, `beatmap_id`, `has_replay`, `preserve`, `rank`, `passed`, `accuracy`, `max_combo`, `total_score`, `data`, `pp`, `legacy_score_id`, `legacy_total_score`, `ended_at`, `unix_updated_at`) VALUES ");
 
             Console.WriteLine($" Processing scores {scores.First().score_id} to {scores.Last().score_id}");
             Stopwatch sw = new Stopwatch();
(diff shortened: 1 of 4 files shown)


osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/BatchInserter.cs
```diff
@@ -87,12 +89,6 @@ private async Task run(HighScore[] scores)
             {
                 try
                 {
-                    if (highScore.score_id == 0)
-                    {
-                        // Something really bad probably happened, abort for safety.
-                        throw new InvalidOperationException("Score arrived with no ID");
-                    }
```
bdach, line 94: Why is this safety removed?

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/WatchScoresCommand.cs
```diff
@@ -0,0 +1,182 @@
...
+
+            outputFinalStats();
+            return 0;
+        }
+
+        private void pushCompletedScoreToQueue(BatchInserter inserter)
+        {
+            if (scoreStatisticsQueueProcessor != null)
+            {
+                Debug.Assert(inserter.ElasticScoreItems.Any());
```
bdach, line 135: I'm very confused here. The check is on `scoreStatisticsQueueProcessor`, but this assert is on `ElasticScoreItems`. First, I would expect this assert to be _the other way around_, i.e.

```suggestion
                Debug.Assert(!inserter.ElasticScoreItems.Any());
```

because either `ElasticScoreItems` or `ScoreStatisticsItems` should be empty, and this is the branch pushing the second of these. But, more importantly, why does anything in here check if there are any items in `ElasticScoreItems`, _if nothing in this command actually pushes anything to ES ever, and arguably should never do so?_

I won't mention the fact that `SkipScoreProcessor` is a rather weird flag in this command, despite the fact that it _kinda_ does what you'd expect it to, but it does so in a very weird way (by appending items to `ElasticScoreItems` which will never be pushed to ES).

peppy: ES indexing is never a thing with non-high scores. I've removed this remaining reference.

bdach (approved):

smoogipoo (approved):
lgtm

PR #208, opened 2024-01-30
Don't throw on failure to convert scores

peppy:
All failed scores are failing right now in osu! mode.

This is just a workaround until we fix the underlying issue.

Tracking at https://github.com/ppy/osu-queue-score-statistics/issues/209

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/BatchInserter.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/BatchInserter.cs
@@ -49,6 +49,7 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Queue
 
         private readonly bool importLegacyPP;
         private readonly bool dryRun;
+        private readonly bool throwOnFailure;
 
         public HighScore[] Scores { get; }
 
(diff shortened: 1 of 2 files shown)


osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/BatchInserter.cs
```diff
@@ -164,7 +166,10 @@ private async Task run(HighScore[] scores)
                 }
                 catch (Exception e)
                 {
-                    throw new AggregateException($"Processing legacy score {highScore.score_id} failed.", e);
+                    if (throwOnFailure)
+                        throw new AggregateException($"Processing legacy score {highScore.score_id} failed.", e);
+
+                    Console.WriteLine($"Processing legacy score {highScore.score_id} failed.", e);
```
bdach, line 172:
```suggestion
                    Console.WriteLine($"Processing legacy score {highScore.score_id} failed: {e}");
```

pretty sure this isn't gonna do what you'd expect it to do otherwise

bdach (changes requested):

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/BatchInserter.cs
peppy: Hmm, in my previous version i was throwing away the exception, which is actually nicer in logging.

So... it does what I want, just in a weird way 😅

bdach: alright well then 
```suggestion
                    Console.WriteLine($"Processing legacy score {highScore.score_id} failed.");
```
i guess

bdach (approved):

PR #211, opened 2024-01-31
Add test coverage of user total performance processor correctly grouping by beatmap

bdach:
This is just test coverage, and just to assuage my paranoia. If nobody else needs this it's fine, it can be closed instantly without further review, I just needed to test things to make sure nothing is broken.

The long and short of why I wrote it is that because we're [not marking scores as non-preserved yet](https://github.com/ppy/osu-queue-score-statistics/issues/141), setting multiple scores on the same beatmap will result in them showing on user profiles i.e. in recent section, with pp point counts shown. So I just wanted to make sure this doesn't somehow result in the pp total being miscounted. And it doesn't, and it's probably obvious that it doesn't from the implementation, but all I'm saying that is something that is difficult to check empirically and I think having this exercised in tests is worthwhile.


peppy (approved):
Good coverage to have, especially as we might change this again in the future 👍

PR #212, opened 2024-01-31
Backfill PP in `ScorePerformanceProcessor` to `osu_scores_high` on legacy scores

### LINKED ISSUE #183: osu!stable score screen needs consideration for the new PP processing
I don't know how to title this issue to be most effective, sorry.

1. `osu-web-10` polls `osu_scores_high` table waiting for the `pp` column to be processed. This column will no longer receive data after osu-performance is taken offline.
2. `osu-web-10` expects `osu_user_stats` to be updated in the same transaction as the `pp` column from the above, as it immediately returns the total PP/rank from there as soon as the `pp` column is populated. `osu-queue-score-statistics` does not run everything in a single transaction.

Technically, we can populate the `pp` column from [`UserTotalPerformanceProcessor`](https://github.com/ppy/osu-queue-score-statistics/blob/master/osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserTotalPerformanceProcessor.cs).  
Note: Currently [`ScorePerformanceProcessor`](https://github.com/ppy/osu-queue-score-statistics/blob/master/osu.Server.Queues.ScoreStatisticsProcessor/Processors/ScorePerformanceProcessor.cs) doesn't update the in-memory `SoloScoreInfo` pp value, and only inserts it into the database. We should probably make it update the model if this is the path forward.

tsunyoku:
Resolves #183.

From reading the issue it seems the end of the processor is the desired place to put this, but I can move it if someone thinks it belongs in a better place (personally I think it should be updated [here](https://github.com/ppy/osu-queue-score-statistics/blob/2e76451d61efe660aa8c6cf6db8b2c195f03ff65/osu.Server.Queues.ScoreStatisticsProcessor/Processors/ScorePerformanceProcessor.cs#L137-L141) but it shouldn't make any difference as it's in a single transaction anyway)

Updates the `SoloScoreInfo` pp value as suggested in the issue.

It would be good if someone could confirm if `SoloScoreInfo.LegacyScoreId` will be `osu_scores_high.score_id` if the score is a pass. If it isn't, then it will require an extra lookup to `osu_scores` to fetch `osu_scores.high_score_id` to be able to update the high score (if it exists).

Once this is in, it *should* allow for osu-performance's pp processor to be turned off for legacy scores.


bdach:
I'd say this needs some sort of configuration flag to allow toggling this write otherwise managing this is gonna get annoying (say, something blows up, and we have to back this change out and bring `osu-performance` back, but the only way to do this would be to revert this change).

Compare the symmetrical change to this (https://github.com/ppy/osu-performance/pull/177) which is behind a config variable.

smoogipoo:
I suggest writing this as a processor itself. Feels ugly sprinkling code into the queue processor, but also that way the env variable isn't needed because it's already present as [`DisabledProcessors`](https://github.com/ppy/osu-queue-score-statistics/blob/185ad3c00423f6209bcfd8f7177fbe915179055a/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/WatchQueueCommand.cs#L14).

osu.Server.Queues.ScoreStatisticsProcessor/ScoreStatisticsQueueProcessor.cs
```diff
@@ -211,6 +220,16 @@ private IEnumerable<IProcessor> enumerateValidProcessors(SoloScoreInfo score)
             return result;
         }
 
+        private static void updateLegacyScorePerformance(SoloScoreInfo score, MySqlConnection db, MySqlTransaction transaction)
+        {
+            string highScoresTable = LegacyDatabaseHelper.GetRulesetSpecifics(score.RulesetID).HighScoreTable;
+            db.Execute($"UPDATE {highScoresTable} SET pp = @Pp WHERE score_id = @LegacyScoreId", new
```
smoogipoo, line 226: For scores imported from `osu_scores` (the non-`high` table), `LegacyScoreId` will be 0.

I'm not sure how to handle this - do these scores need to be updated? @peppy

smoogipoo (changes requested):

osu.Server.Queues.ScoreStatisticsProcessor/ScoreStatisticsQueueProcessor.cs
tsunyoku: Yeah, that was the clarification I needed. I think a `score.IsLegacyScore && score.LegacyScoreId > 0` should suffice here

tsunyoku:
I've moved it into it's own processor but it leaves me with one question - do the performance related commands which run the `ScorePerformanceProcessor` need updating to run this processor too? As I understand the column is only used to display on the charts via stable rather than actually being used somewhere else (i.e the API) but it would probably still be good to prevent those tables having "out of date" or "incorrect" values. I can update if necessary.

bdach:
I'm not sure I know the answer to that question to be honest, but I'd say "probably".

smoogipoo:
Here's one such run: https://github.com/ppy/osu/actions/runs/7726219762 Maybe there's some data that can be extracted from it.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/LegacyScorePerformanceProcessor.cs
```diff
@@ -0,0 +1,59 @@
...
+        public void ApplyGlobal(SoloScoreInfo score, MySqlConnection conn)
+        {
+        }
+
+        public async Task ProcessScoreAsync(SoloScoreInfo score, MySqlConnection connection, MySqlTransaction? transaction)
+        {
+            // Processor should only be ran on legacy high scores.
+            // `score.LegacyScoreId == 0` check is required as the ID will be 0 for non-`high` scores.
+            // `score.Passed` is usually checked via "RunOnFailedScores", but this method is also used by the CLI batch processor.
+            if (!score.IsLegacyScore || score.LegacyScoreId == 0 || !score.Passed)
```
smoogipoo, line 48: At the end of the day this simplifies to `if (score.LegacyScoreId == null || score.LegacyScoreId == 0 || ...)`. Which make me question - can this method just use `SoloScore` as is and avoid the `.ToScoreInfo()` from above?

smoogipoo:
@peppy is there anything guaranteeing this processor runs after the _legacy_ osu-performance PP calculation when a score is submitted?

peppy:
> @peppy is there anything guaranteeing this processor runs after the _legacy_ osu-performance PP calculation when a score is submitted? The concern here would be for this processor to update PP on legacy scores, and then osu-performance runs to set it back to an old version.

Yes, currently it is guaranteed, as the processor would only run after the score arrived in the new table. Scores only arrive in the new table after the `score_process_queue` hits a state where PP has been populated (this is a requirement to ensure the pp value is filled in the new table).

smoogipoo (approved):
Lgtm

smoogipoo:
@peppy will leave the final review/merge to you since you requested a review.

peppy:
Looping back to this one, is there a reason we can't implement this as a very simple:

```diff
diff --git a/osu.Server.Queues.ScoreStatisticsProcessor/Processors/ScorePerformanceProcessor.cs b/osu.Server.Queues.ScoreStatisticsProcessor/Processors/ScorePerformanceProcessor.cs
index 9faebb2..0e165c8 100644
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Processors/ScorePerformanceProcessor.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Processors/ScorePerformanceProcessor.cs
@@ -158,6 +158,15 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Processors
                     ScoreId = score.ID,
                     Pp = performanceAttributes.Total
                 }, transaction: transaction);
+
+                if (score.IsLegacyScore)
+                {
+                    var helper = LegacyDatabaseHelper.GetRulesetSpecifics(score.RulesetID);
+                    await connection.ExecuteAsync($"UPDATE {helper.HighScoreTable} SET pp = @Pp WHERE score_id = @LegacyScoreId", new
+                    {
+                        LegacyScoreId = score.LegacyScoreId,
+                    }, transaction: transaction);
+                }
             }
             catch (Exception ex)
             {

```

?

tsunyoku:
That is what it was initially (see https://github.com/ppy/osu-queue-score-statistics/pull/212/commits/7532dc6e0ee1a3dc3602a3c4ec2588876364f9e1). This and a setting to disable writing legacy pp should be sufficient, but smoogi felt weird about sprinkling it into that processor so I moved it into it's own.

smoogipoo:
I don't feel strongly about it, go ahead.

bdach:
Don't feel strongly either but do note that with the proposed patch, if this has to be reverted for whatever reason it'll need to be a full revert + redeploy rather than just an envvar tweak. If it's all the same to you then I'm completely hands off.

tsunyoku:
I'll revert back to doing the update inside the normal performance processor, but I'll also re-add the environment variable so that it can be toggled without a redeploy

peppy (approved):

peppy:
Importantly, the processor is currently disabled for legacy scores via

https://github.com/ppy/osu-queue-score-statistics/blob/9afb1b10c612654c439d87e11b28aa0d26386b95/osu.Server.Queues.ScoreStatisticsProcessor/Processors/ScorePerformanceProcessor.cs#L37

so this change will have no immediate effect if deployed.

PR #214, opened 2024-02-01
Order by `score_id` instead of `queue_id`

peppy:
Probably never going to actually be an issue, but this is more correct. Basically as there's no transaction around the score and queue insert during web-10 score submission, two submissions could execute in a different ordering. We should preserve `score_id` ordering as much as possible.


osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/WatchHighScoresCommand.cs
```diff
@@ -133,7 +133,7 @@ public async Task<int> OnExecuteAsync(CancellationToken cancellationToken)
                                                  }))
                                              // there might be multiple queue entries for the same insert. this can cause issues due to how we do the mapping lookup so let's fix that.
                                              .DistinctBy(s => s.score_id)
-                                             .OrderBy(s => s.queue_id)
+                                             .OrderBy(s => s.score)
```
bdach, line 136: uhhhhhhhhhhhhhhhhhhhhhhhhhhh

```suggestion
                                             .OrderBy(s => s.score_id)
```

bdach (changes requested):

bdach (approved):

PR #215, opened 2024-02-01
Add a timeout condition to avoid score watchers getting stuck

peppy:
This has happened twice now, where a score in the queue table ends up stuck in a `status=0` state.

I'm not yet sure why this happens, but given how rare it is my primary goal is to just keep things running so they don't need manual intervention. (and we already have verification procedures to amend incorrect pp should it actually be visible to a user).

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/WatchHighScoresCommand.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/WatchHighScoresCommand.cs
@@ -163,7 +163,7 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Queue
                             Console.WriteLine($"Importing score which was not processed due to timeout (score_id: {pendingProcessing.score_id} queue_id: {pendingProcessing.queue_id})");
 
                             // only process up to the timed-out entry.
-                            highScores = highScores.Where(s => s.queue_id <= pendingProcessing.score_id).ToArray();
+                            highScores = highScores.Where(s => s.queue_id <= pendingProcessing.queue_id).ToArray();
                         }
                         else
                         {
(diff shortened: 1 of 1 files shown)


osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/WatchHighScoresCommand.cs
```diff
@@ -136,18 +140,44 @@ public async Task<int> OnExecuteAsync(CancellationToken cancellationToken)
...
+                            Console.WriteLine($"Waiting on processing of (score_id: {pendingProcessing.score_id} queue_id: {pendingProcessing.queue_id})");
+                            Thread.Sleep(sleep);
+                            continue;
+                        }
+
+                        // 5 minute timeout
+                        if (pendingProcessingWaitTime > 300000)
+                        {
+                            Console.WriteLine($"Importing score which was not processed due to timeout (score_id: {pendingProcessing.score_id} queue_id: {pendingProcessing.queue_id})");
+                        }
```
bdach, line 164: Isn't this going to import not only the score with id `pendingProcesing.score_id`, but the rest of the involved batch as well, which may contain other unprocessed scores? Should that be logged at least?

peppy: Sounds fair to only process up to the currently pending item and continue afterwards for any furthers.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/WatchHighScoresCommand.cs
```diff
@@ -136,18 +140,47 @@ public async Task<int> OnExecuteAsync(CancellationToken cancellationToken)
...
+                            continue;
+                        }
+
+                        // 5 minute timeout
+                        if (pendingProcessingWaitTime > 300000)
+                        {
+                            Console.WriteLine($"Importing score which was not processed due to timeout (score_id: {pendingProcessing.score_id} queue_id: {pendingProcessing.queue_id})");
+
+                            // only process up to the timed-out entry.
+                            highScores = highScores.Where(s => s.queue_id <= pendingProcessing.score_id).ToArray();
```
bdach, line 166: Uh... isn't this mixing two disparate ID schemes?

```suggestion
                            highScores = highScores.Where(s => s.queue_id <= pendingProcessing.queue_id).ToArray();
```

Also doing things this way means that if legacy processing just dies this is going to throttle everything to a rate of 1 score every 5 minutes. Not sure if that's a problem.

bdach (changes requested):

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/WatchHighScoresCommand.cs
peppy: The alerts we have setup require higher than that rate, so I think it should be okay. And is probably for the best if more than the odd entry are not being processed for PP.

bdach (approved):

PR #217, opened 2024-02-01
Fix `BatchInserter` not showing correct process range due to no guaranteed ordering

peppy:
Noticed during last run. Maybe adds small overhead but probably fine.


bdach:
https://github.com/ppy/osu-queue-score-statistics/pull/214 should enforce this already for the high case but I guess the importer is also used for non-high so probably fine.

bdach (approved):

PR #220, opened 2024-02-08
Change verify command to support batched writes

peppy:
I'd been meaning to do this for a while now, in order to allow rewriting ranks to the correct equations, something I've not been able to do due to the overhead of individual SQL writes in this command.

I already have this stabilised and running on production, updating around 3,000 rows per second (ample for this kind of operation).

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/VerifyImportedScoresCommand.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/VerifyImportedScoresCommand.cs
@@ -10,10 +10,12 @@ using System.Threading;
 using System.Threading.Tasks;
 using Dapper;
 using McMaster.Extensions.CommandLineUtils;
+using MySqlConnector;
 using osu.Game.Scoring;
 using osu.Server.QueueProcessor;
 using osu.Server.Queues.ScoreStatisticsProcessor.Commands.Queue;
 using osu.Server.Queues.ScoreStatisticsProcessor.Helpers;
+using StringBuilder = System.Text.StringBuilder;
 
 namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Maintenance
 {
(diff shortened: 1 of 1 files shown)


osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/VerifyImportedScoresCommand.cs
```diff
@@ -286,31 +238,54 @@ public async Task<int> OnExecuteAsync(CancellationToken cancellationToken)
...
+                    Console.WriteLine();
+                    Console.WriteLine($"Flushing sql batch ({bufferLength:N0} bytes)");
+                    conn.Execute(sqlBuffer.ToString());
+
+                    if (elasticItems.Count > 0)
+                    {
+                        if (!DryRun)
+                            elasticQueueProcessor.PushToQueue(elasticItems.ToList());
+                        Console.WriteLine($"Queued {elasticItems.Count} items for indexing");
+                    }
```
bdach, line 273: Pushing to elastic here is potentially wrong. This is because `elasticItems` is not cleared here; it is cleared in the main loop. So there is a possibility of the following happening:

1. Batch 1 is verified. Items are added to `elasticItems` for elastic processing as this is happening.
2. Batch 1 is flushed. If the accumulated SQL buffer length is below 1024 bytes, none of this executes, so the items aren't pushed for processing to elastic.
3. Batch 2 is verified. At this point `elasticItems` is cleared in the main processing loop. The items in that list cannot be recovered again and will never be pushed to elastic even when the relevant buffered SQL statements from step (2) execute.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/VerifyImportedScoresCommand.cs
```diff
@@ -286,31 +238,54 @@ public async Task<int> OnExecuteAsync(CancellationToken cancellationToken)
...
+            {
+                if (!DryRun)
+                {
+                    Console.WriteLine();
+                    Console.WriteLine($"Flushing sql batch ({bufferLength:N0} bytes)");
+                    conn.Execute(sqlBuffer.ToString());
+
+                    if (elasticItems.Count > 0)
+                    {
+                        if (!DryRun)
```
bdach, line 270: check is redundant (already is in an enclosing `!DryRun` if block).

bdach (changes requested):

bdach:
Please double-check https://github.com/ppy/osu-queue-score-statistics/pull/220/commits/65af7e88d893add0e0fea64101fd59728ed8cc54; looks okay otherwise.

bdach (approved):

PR #224, opened 2024-02-12
Fix `BatchInserter` optimisation breaking insertion order requirements

peppy:
As discussed in internal discord. 

Fixes the underlying cause of https://github.com/ppy/osu-web/issues/10922.


bdach (approved):

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/BatchInserter.cs
```diff
@@ -182,10 +169,24 @@ private async Task run(HighScore[] scores)
...
 
+            StringBuilder insertBuilder = new StringBuilder(
+                "INSERT INTO scores (`user_id`, `ruleset_id`, `beatmap_id`, `has_replay`, `preserve`, `ranked`, `rank`, `passed`, `accuracy`, `max_combo`, `total_score`, `data`, `pp`, `legacy_score_id`, `legacy_total_score`, `ended_at`, `unix_updated_at`) VALUES ");
+
+            foreach (var score in scores)
+            {
+                if (string.IsNullOrEmpty(score.InsertSql))
+                    continue;
+
+                if (!first)
```
smoogipoo, line 183: I had to scroll up to the veeeeeeeeeeeeeeeeeeery start of this method, through parallelised code, to find this variable. Please move it closer to here in your next changes so I can be more assured that there isn't some weird threading thing going on.

PR #225, opened 2024-02-19
Add command to verify presence of replays on S3

peppy:
A simple one. To be used hopefully once-ever, to fix https://github.com/orgs/ppy/projects/19/views/1?pane=issue&itemId=51932194.

https://github.com/orgs/ppy/projects/19/views/1?pane=issue&itemId=51932194

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/VerifyReplayExistsCommand.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/VerifyReplayExistsCommand.cs
@@ -1,0 +1,108 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Amazon;
+using Amazon.Runtime;
+using Amazon.S3;
+using Dapper;
+using McMaster.Extensions.CommandLineUtils;
+using osu.Server.QueueProcessor;
+using osu.Server.Queues.ScoreStatisticsProcessor.Models;
+
+namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Maintenance
+{
+    [Command("verify-replays-exist", Description = "Verifies replays exist on S3 where replay flag implies they do.")]
+    public class VerifyReplaysExistCommand
+    {
+        /// <summary>
+        /// The high score ID to start deleting imported high scores from.
+        /// </summary>
+        [Option(CommandOptionType.SingleValue, Template = "--start-id")]
+        public ulong? StartId { get; set; }
+
+        [Option(CommandOptionType.SingleOrNoValue, Template = "-v|--verbose", Description = "Output when a score is preserved too.")]
+        public bool Verbose { get; set; }
+
+        /// <summary>
+        /// The number of scores to run in each batch. Setting this higher will cause larger SQL statements for insert.
+        /// </summary>
+        [Option(CommandOptionType.SingleValue, Template = "--batch-size")]
+        public int BatchSize { get; set; } = 5000;
+
+        [Option(CommandOptionType.SingleOrNoValue, Template = "--dry-run")]
+        public bool DryRun { get; set; }
+
+        public async Task<int> OnExecuteAsync(CancellationToken cancellationToken)
+        {
+            ulong lastId = StartId ?? 0;
+
+            var s3Key = Environment.GetEnvironmentVariable("S3_KEY") ?? string.Empty;
+            var s3Secret = Environment.GetEnvironmentVariable("S3_SECRET") ?? string.Empty;
+
+            using var s3 = new AmazonS3Client(new BasicAWSCredentials(s3Key, s3Secret), new AmazonS3Config
+            {
+                CacheHttpClient = true,
+                HttpClientCacheSize = 32,
+                RegionEndpoint = RegionEndpoint.USWest1,
+                UseHttp = true,
+                ForcePathStyle = true
+            });
+
+            using var conn = DatabaseAccess.GetConnection();
+
+            Console.WriteLine();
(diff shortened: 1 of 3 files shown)


osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/VerifyReplayExistsCommand.cs
```diff
@@ -0,0 +1,108 @@
...
+
+            var s3Key = Environment.GetEnvironmentVariable("S3_KEY") ?? string.Empty;
+            var s3Secret = Environment.GetEnvironmentVariable("S3_SECRET") ?? string.Empty;
+
+            using var s3 = new AmazonS3Client(new BasicAWSCredentials(s3Key, s3Secret), new AmazonS3Config
+            {
+                CacheHttpClient = true,
+                HttpClientCacheSize = 32,
+                RegionEndpoint = RegionEndpoint.USWest1,
+                UseHttp = true,
```
bdach, line 53: http (as opposed to https, I imagine, although the docs are [crap](https://docs.aws.amazon.com/sdkfornet/v3/apidocs/items/Runtime/TClientConfig.html))? Sure about that?

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/VerifyReplayExistsCommand.cs
```diff
@@ -0,0 +1,108 @@
...
+
+                if (!importedScores.Any())
+                {
+                    if (lastId > await conn.QuerySingleAsync<ulong>("SELECT MAX(id) FROM scores"))
+                    {
+                        Console.WriteLine("All done!");
+                        break;
+                    }
+
+                    lastId += (ulong)BatchSize;
```
bdach, line 83: Why is `lastId` incremented only in this branch? Doesn't this mean that this is going to spin over the first batch of scores forevermore?

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/VerifyReplayExistsCommand.cs
```diff
@@ -0,0 +1,108 @@
...
+            Console.WriteLine();
+            Console.WriteLine($"Verifying replay flag for scores starting from {lastId}");
+
+            if (DryRun)
+                Console.WriteLine("RUNNING IN DRY RUN MODE.");
+
+            while (!cancellationToken.IsCancellationRequested)
+            {
+                IEnumerable<SoloScore> importedScores = await conn.QueryAsync<SoloScore>(
+                    "SELECT `id`, `has_replay` FROM scores WHERE id BETWEEN @lastId AND (@lastId + @batchSize - 1) AND has_replay = 1 AND legacy_score_id IS NULL ORDER BY id",
```
bdach, line 68: Why select `has_replay` if the filter already specifies `has_replay = 1`?

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/VerifyReplayExistsCommand.cs
```diff
@@ -0,0 +1,108 @@
...
+                        break;
+                    }
+
+                    lastId += (ulong)BatchSize;
+                    continue;
+                }
+
+                foreach (var score in importedScores)
+                {
+                    if (!score.has_replay) continue;
```
bdach, line 89: In conjunction with the above, why is this being checked at all? `has_replay` should always be true making this a dead branch.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/VerifyReplayExistsCommand.cs
```diff
@@ -0,0 +1,108 @@
...
+
+                    try
+                    {
+                        // check the replay is available on s3.
+                        _ = await s3.GetObjectMetadataAsync("score-replays", score.id.ToString(CultureInfo.InvariantCulture), cancellationToken);
```
bdach, line 94: As already touched on above, this feels quite dodge. Any old network failure and this deletes replays. No error logging either. Is there really no better way to at least check for a slightly more specific error?

There's [some old code online that does this](https://github.com/aws/aws-sdk-net/blob/c688adc3850fbcbf708154258963afbb680b59d1/sdk/src/Services/S3/Custom/_bcl/IO/S3FileInfo.cs#L104-L147), but then again, not sure how up-to-date that is, given `S3FileInfo.Exists()` was apparently a method offered out of the box in AWS SDK, but now it doesn't seem to be there at all anymore...

bdach (changes requested):

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/VerifyReplayExistsCommand.cs
peppy: This was just copypaste from `osu-server-spectator`. I believe there was probably a reason for this (cloudflare r2 support maybe?) at some point.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/VerifyReplayExistsCommand.cs
peppy: This was also just copypaste from our S3 helper utility in `osu-server-spectator`. But I agree it's pretty dodge.

peppy: I'll give what you linked a try, but the exception being thown is `PermissionDenied`, which is the same for a multitude of cases including `NotFound`. I'm not sure we get any more fidelity in exceptions these days from S3.

bdach: Sure, I'd still say matching on that is better than indiscriminate catching.

bdach (approved):
Seems ok

PR #227, opened 2024-02-27
Stop marking scores as `ranked = 0` for pp only purposes

peppy:
We use this flag for cases where scores should not be considered for PP or leaderboards now. Cases where PP is not required simply leaves it as null.


smoogipoo (approved):

bdach:
> Cases where PP is not required simply leaves it as null.

I assume this is going to regress the pp display fix made at https://github.com/ppy/osu/pull/27053 for scores set e.g. on unranked mods? cc @ppy/team-web

peppy:
Yeah, good call there. We'll either need to have local logic to check the conditions for whether a score should award PP, or come up with another value to represent what we were using `null` for..

PR #228, opened 2024-03-04
Update game packages

bdach:
I'm doing this so that https://github.com/ppy/osu/pull/27361 can be deployed live and broken leaderboard can be fixed.


bdach:
Blocking because https://github.com/ppy/osu/issues/27496 might need another scoring algo fix, another tag, another bump etc etc.

peppy (approved):

PR #229, opened 2024-03-04
Fix play count processor not incrementing pass count

bdach:
Intends to fix https://github.com/ppy/osu/issues/27485 for new incoming scores. I don't know what happens to existing scores. Is there a tool that can recompute these already maybe?

As far as I can tell the implementation of this should match `osu-web-10`. It makes *some* sense to me when I squint at it, but I'm not super sure.


smoogipoo (approved):
Matches as far as I can tell but I'm not going to merge this in case @peppy has thoughts on retroactively applying it.

peppy (approved):

PR #230, opened 2024-03-08
Prepare infrastructure for implementing hush-hush medals

bdach:
- [x] Depends on https://github.com/ppy/osu-queue-score-statistics/pull/195 for mergeability

RFC. PRing so that the thing I've spent multiple days on now can be looked at closely and accurately without me typing up hand-wavey essays. Will want input from @peppy and @ThePooN.

## How this is supposed to work

The premise is that this project is to remain public, but will consume a private DLL where the hush-hush medals are implemented.

In principle the setup I'm about to describe requires no new nuget packages, but will require separate docker images. The loose sketch of how I imagine the private part is going to look is available for preview at https://github.com/bdach/osu-hush-hush-medals-poc. (The project is intended to be public, it contains no actual code that needs to be kept secret. It's just a PoC.)

This works as follows:

1. The private project consumes the public project at a _source level_ as a submodule.
2. The private project references the public project at a _.NET level_, which means it can consume all of its public API.
3. The public project remains the primary runnable one, but it can load private DLLs with private processors and medal awarders via the `EXTERNAL_PROCESSOR_PATH` envvar.

    I know what you're probably thinking: this means that there is a circular dependency now. Please read on despite this.
4. The private project contains a Dockerfile. That Dockerfile builds an image from the private project, and thus transitively builds the public project too as the private one depends on it. Thus the build output contains both the private and public DLLs, as output by _a single compilation_.

    This is important because it is _transparent_ (as in, you can deal with deployment of only the private version) and _fail-safe_. Yes the proper way to do this would be to move `IProcessor`, `IMedalAwarder`, etc. out to a separate "models" nuget and have the public and private projects reference that, and I initially did this. But that makes it possible for a _binary incompatibility_ of the shared models subproject if you forget to bump one or the other at either end, which then will most likely die at runtime because we're still going to be wanting to load the private DLL. If the two are built together via the transitive dependency, and deployed together, this is _statically_ avoided. If anything breaks, the project just won't build.
5. The private docker image will be the one that will be deployed. Its entrypoint is the public DLL, with the private DLL loaded in as pointed by the envvar.

As part of review of this, please scrutinise the following in the PoC too:
- [`MyFirstMedalAwarder`](https://github.com/bdach/osu-hush-hush-medals-poc/blob/master/osu.Server.Queues.ScoreStatisticsProcessor.HushHushMedals/MyFirstMedalAwarder.cs)
- [`MyFirstMedalAwarderTest`](https://github.com/bdach/osu-hush-hush-medals-poc/blob/master/osu.Server.Queues.ScoreStatisticsProcessor.HushHushMedals.Tests/MyFirstMedalAwarderTest.cs)
- [`Dockerfile`](https://github.com/bdach/osu-hush-hush-medals-poc/blob/master/Dockerfile)

## Testing

The PoC above can be tested full-stack with a local osu-web checkout setup.

1. Clone https://github.com/bdach/osu-hush-hush-medals-poc.
2. Ensure that the LIO endpoint is pointed at localhost (in the _submodule_ of the poc) and that `SHARED_INTEROP_SECRET` is set web-side.
3. Change [`MyFirstMedalAwarder`](https://github.com/bdach/osu-hush-hush-medals-poc/blob/master/osu.Server.Queues.ScoreStatisticsProcessor.HushHushMedals/MyFirstMedalAwarder.cs#L16) to point at a valid medal ID in your db. (It'll attempt to grant the medal after every completed play until it succeeds. If it succeeds and you wanna try again without rebuilding, just truncate `osu_user_achievements`.)
4. Build the image:
    ```sh
    $ docker build . -t osu-queue-score-statistics-private
    ```
5. Run the image:
    ```sh
    $ docker run --rm --net=host -e 'SHARED_INTEROP_SECRET=${your-secret}' osu-queue-score-statistics-private queue watch
    ```
    (`--net=host` is used so that the container can see the ports that web docker compose forwarded out. I'm not sure if there's a better way to do this.)
6. Launch game and finish any beatmap. You should see the medal you chose in point (3) be awarded.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,206 @@
+# EditorConfig is awesome: http://editorconfig.org
+root = true
+
+[*.{csproj,props,targets}]
+charset = utf-8-bom
+end_of_line = crlf
+insert_final_newline = true
+indent_style = space
+indent_size = 2
+trim_trailing_whitespace = true
+
+[g_*.cs]
+generated_code = true
+
+[*.cs]
+end_of_line = crlf
+insert_final_newline = true
+indent_style = space
+indent_size = 4
+trim_trailing_whitespace = true
+
+#license header
+file_header_template = Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.\nSee the LICENCE file in the repository root for full licence text.
+
+#Roslyn naming styles
+
+#PascalCase for public and protected members
+dotnet_naming_style.pascalcase.capitalization = pascal_case
+dotnet_naming_symbols.public_members.applicable_accessibilities = public,internal,protected,protected_internal,private_protected
+dotnet_naming_symbols.public_members.applicable_kinds = property,method,field,event
+dotnet_naming_rule.public_members_pascalcase.severity = error
+dotnet_naming_rule.public_members_pascalcase.symbols = public_members
+dotnet_naming_rule.public_members_pascalcase.style = pascalcase
+
+#camelCase for private members
+dotnet_naming_style.camelcase.capitalization = camel_case
+
+dotnet_naming_symbols.private_members.applicable_accessibilities = private
+dotnet_naming_symbols.private_members.applicable_kinds = property,method,field,event
+dotnet_naming_rule.private_members_camelcase.severity = warning
+dotnet_naming_rule.private_members_camelcase.symbols = private_members
+dotnet_naming_rule.private_members_camelcase.style = camelcase
+
+dotnet_naming_symbols.local_function.applicable_kinds = local_function
+dotnet_naming_rule.local_function_camelcase.severity = warning
+dotnet_naming_rule.local_function_camelcase.symbols = local_function
+dotnet_naming_rule.local_function_camelcase.style = camelcase
+
+#all_lower for private and local constants/static readonlys
+dotnet_naming_style.all_lower.capitalization = all_lower
+dotnet_naming_style.all_lower.word_separator = _
+
+dotnet_naming_symbols.private_constants.applicable_accessibilities = private
+dotnet_naming_symbols.private_constants.required_modifiers = const
+dotnet_naming_symbols.private_constants.applicable_kinds = field
+dotnet_naming_rule.private_const_all_lower.severity = warning
+dotnet_naming_rule.private_const_all_lower.symbols = private_constants
+dotnet_naming_rule.private_const_all_lower.style = all_lower
+
+dotnet_naming_symbols.private_static_readonly.applicable_accessibilities = private
+dotnet_naming_symbols.private_static_readonly.required_modifiers = static,readonly
+dotnet_naming_symbols.private_static_readonly.applicable_kinds = field
+dotnet_naming_rule.private_static_readonly_all_lower.severity = warning
+dotnet_naming_rule.private_static_readonly_all_lower.symbols = private_static_readonly
+dotnet_naming_rule.private_static_readonly_all_lower.style = all_lower
+
+dotnet_naming_symbols.local_constants.applicable_kinds = local
+dotnet_naming_symbols.local_constants.required_modifiers = const
+dotnet_naming_rule.local_const_all_lower.severity = warning
+dotnet_naming_rule.local_const_all_lower.symbols = local_constants
+dotnet_naming_rule.local_const_all_lower.style = all_lower
+
+#ALL_UPPER for non private constants/static readonlys
+dotnet_naming_style.all_upper.capitalization = all_upper
+dotnet_naming_style.all_upper.word_separator = _
+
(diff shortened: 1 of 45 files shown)


ThePooN:
I don't think this is the most elegant solution, but there wouldn't be a clear benefit doing otherwise, so it doesn't matter to me.

Once this is merged, I propose adding a GitHub workflow on the private repo that would update the submodule on every tag from the public repo, then execute tests, and commit if they pass. That commit would then trigger the private image build in preparation for deployment.  
The current workflow on the public repo would also stay for public use and staging.

bdach:
> I don't think this is the most elegant solution

I'm open to exploring alternative solutions if you think there are better ways. I was attempting to optimise for ease of deployment first and foremost.

ThePooN:
Well, I'm referring to the discussion we've already had on this topic. The way I'd have done it is publishing osu-queue-score-statistics on NuGet, and making this private project have its own entrypoint that would set-up and run osu-queue-score-statistics with extra medal processors enabled. Or the radically different approach of creating another queue and queue processor dedicated to hush-hush medals that would be completely independent of osu-queue-score-statistics.

On my end, there's only a single project to deploy regardless, and I can also make sense and automate the process you already laid out. I feel the decision here depends more on the maintenance costs and future developments which falls on you both, so I'm pretty indifferent.

bdach:
> publishing osu-queue-score-statistics on NuGet

That can be done with very little adjustment to what this PR proposes if it's not considered too much process. Interested in what @peppy thinks about that.

> creating another queue and queue processor dedicated to hush-hush medals that would be completely independent of osu-queue-score-statistics

As previously discussed I think this would be too much ceremony and too many new moving pieces that can fail.

substanc3-dev:
> Or the radically different approach of creating another queue and queue processor dedicated to hush-hush medals that would be completely independent of osu-queue-score-statistics.

I think it may be desirable to extend this project (and also server-spectator) in more ways than just the hush-hush medals in the private layer in the long run, so going for a more versatile approach like this (or publishing on nuget with custom bootstrap as suggested) probably makes more sense.

bdach:
If we'll need it we can make it happen eventually. But for now it's not worth it imo. The private part is data-dependent on the public part (need other stats updates to take place first). Which means another redis queue, another processor to monitor, etc. It's overdoing it imo. It's architecture astronaucy.

substanc3-dev:
Clarified my comment, was phrased a bit incorrectly, sorry.

peppy:
I've gone through everything here and it all looks good. Not sure if you want to move this out of draft and go for merge yet, but I think it would be fine to do so.

bdach:
Short term I can move out of draft, but would also be interested in your thoughts on the discussion above on how the public side should be consumed by the private side (git submodule or nuget) as it'll give me info on how to proceed further from here.

peppy:
I think submodule is fine as long as the building is automated to automatically pull the latest upstream HEAD and run everything in line with the existing tag flow (thus requiring no manual intervention).

The end goal should be to consume a docker image from a private (likely github) source, rather than the public `osu-queue-score-statistics`, but no extra process added for this to happen.

bdach:
I've updated the private part PoC with two github actions workflows:

## Continuous integration

- [Workflow definition](https://github.com/bdach/osu-hush-hush-medals-poc/blob/master/.github/workflows/ci.yml)
- [Example run](https://github.com/bdach/osu-hush-hush-medals-poc/actions/runs/8282073086)

This isn't much to talk about, most of this is copy-pasted.

## Publish new version

- [Workflow definition](https://github.com/bdach/osu-hush-hush-medals-poc/blob/master/.github/workflows/pack.yml)
- [Example run](https://github.com/bdach/osu-hush-hush-medals-poc/actions/runs/8282079702)

The way this works is slightly different than requested: this workflow doesn't work on tag, but on manual dispatch instead. This is because I want the submodule update & commit to happen _before_ the tag, so pushing a tag kinda doesn't work unless some gymnastics are done. I can do them, but as it turns out you can manual-dispatch workflows via

```sh
$ gh workflow run pack.yml -f tag=2024.314.1
```

in the repo root so maybe fine?

The workflow updates the submodule to latest, runs tests from both projects (this doubles as a compile check), then commits and pushes the submodule update and then the desired tag.

To make the commit https://github.com/stefanzweifel/git-auto-commit-action is used. It looked like a relatively sane action (many users, active, code is mostly just a very sane auditable shell script). Using this workflow does however require a `write:contents` github token to be used. Using nuget would lead to extra steps but eliminate this. Whether that is an acceptable tradeoff is to the eye of the beholder I suppose.

The only parts missing in this workflow is pushing to docker, but that should be a copy-paste affair as well.

---

I think if all of the above sounds fine then the next steps would be to get this in and then transfer the PoC repo (maybe with some slight history rewriting) to the ppy org so that it can be privated, and continue implementation from that point onwards. That said I'm still 100% open to shuffling the process around here still.

peppy (approved):
Can't fault this.

PR #232, opened 2024-03-11
Fix failed plays being included in pack medal query

bdach:
The fact that *this* awarder is the one with the most issues despite being in source for months now is ironic.

As far as I can tell `passed = 1` is the only viable condition here? `ranked = 0` seems useless although I'm not sure what that flag means anymore (I think its goal is now to be set to 0 when a map exits qualified, and most maps in packs are ranked already?) and I'm not sure `preserved = 1` is doing anything right now?

I'm not sure if the indices are able to support this query. Probably needs a safety check.


bdach:
Added `preserve = 1` upon internal counsel. See inline comment and/or commit message for caveats involved.

PR #233, opened 2024-03-12
Exclude more mods from challenge pack medals

cl8n:
...using `IsDifficultyReductionMod()`. this is intended to fix that you can cheese challenge packs with mods like DA, magnetize, relax, etc

the mod check can't be easily translated to SQL, so it's now required to query all of the user's passed scores on the relevant maps, and I can't tell if that is an acceptable tradeoff.

i'm new to this project and might-as-well-be-new to C# so feel free to just edit this if it's easier than reviewing

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Processors/MedalAwarders/PackMedalAwarder.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Processors/MedalAwarders/PackMedalAwarder.cs
@@ -5,6 +5,7 @@ using System.Collections.Generic;
 using System.Linq;
 using Dapper;
 using JetBrains.Annotations;
+using osu.Server.Queues.ScoreStatisticsProcessor.Helpers;
 using osu.Server.Queues.ScoreStatisticsProcessor.Models;
 
 namespace osu.Server.Queues.ScoreStatisticsProcessor.Processors.MedalAwarders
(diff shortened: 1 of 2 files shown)


osu.Server.Queues.ScoreStatisticsProcessor/Processors/MedalAwarders/PackMedalAwarder.cs
```diff
@@ -267,17 +268,31 @@ bool checkPack(int packId, bool noReductionMods = false)
...
-                    + $"WHERE s.user_id = {context.Score.UserID} AND s.passed = 1 AND s.preserve = 1 AND pack_id = {packId} {rulesetCriteria} {modsCriteria}", transaction: context.Transaction);
+                int completed;
+
+                if (noReductionMods)
+                {
+                    var scoresWithBeatmapsetId = context.Connection.Query(
+                        $"SELECT s.data, p.beatmapset_id {selectClauses}",
+                        (SoloScore score, uint beatmapsetId) => (score, beatmapsetId),
+                        transaction: context.Transaction,
+                        splitOn: "beatmapset_id");
```
bdach, line 282: I don't have production access to check empirically but I am expecting that pulling all user scores like that is going to be an issue due to the volume of data involved.

bdach: This is also gonna be a problem [here](https://github.com/ppy/osu-web/blob/b29eeff28e100bd5ed43a26f355575f19918e556/app/Models/BeatmapPack.php#L118-L120).

peppy: I'd feel much more comfortable with the pp based solution for now. As a separate PR which only does that. This can be left in draft for future consideration.

Goal would be to fix this immediately and dwell on what this PR is attempting to do. I do not feel comfortable even beginning to investigate the performance concerns of this PR at this stage.

peppy: can't you just set the pp manually in the test?

bdach: https://github.com/ppy/osu-queue-score-statistics/pull/234

bdach: One corollary of that temporary change is that loved packs won't work for the time being (because loved scores don't get pp). It is what it is for now I guess...

cl8n: thx :pray: 

for loved, there are currently no challenge packs, I'll keep this in mind if it ever comes up

cl8n:
closing since #234 works for now -- in the future, if new difficulty-reducing mods or mod settings are ranked, then the query will have to be updated or something like this PR will have to be revisited.

PR #234, opened 2024-03-12
Do not grant challenge pack medals for scores that don't grant pp

bdach:
Temporary alternative to https://github.com/ppy/osu-queue-score-statistics/pull/233


peppy:
Could also argue we want a `ranked=1` check here, but that's less of a concern and we can revisit later.

peppy (approved):

bdach:
> Could also argue we want a `ranked=1` check here

I don't think it matters unless we start putting qualified maps in packs (because I think that's mostly what the flag is for these days?)

PR #235, opened 2024-03-12
Implement rank milestone medals

bdach:
Closes https://github.com/ppy/osu/issues/27581.

I did mention this wasn't implemented in https://github.com/ppy/osu-queue-score-statistics/pull/195#issuecomment-1944403240, but it was a long PR thread so it might have gotten missed.

~~Not sure if this should be merged imminently while the yo-yo effect with ranks shifting because of the bonus pp discrepancy is still ongoing.~~ I believe this should no longer be a concern.


peppy:
> Not sure if this should be merged imminently while the yo-yo effect with ranks shifting because of the bonus pp discrepancy is still ongoing.

I hope to fix that this week.

peppy (approved):

PR #236, opened 2024-03-13
Delete score pins when cleaning up old user high scores during high score import

peppy:
Score pins generally mark scores as `preserve=1` and therefore can be pinned indefinitely, but for the time being, the legacy score importer is forcefully (hard) deleting scores when they are beaten so that the stored scores match the old tables 1:1. So, let's nuke pinned scores along with the scores themselves to keep parity.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/ImportHighScoresCommand.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/ImportHighScoresCommand.cs
@@ -116,7 +116,7 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Queue
                         await checkSlaveLatency(db, cancellationToken);
 
                     Console.WriteLine($"Fetching next scores from {lastProcessedId}...");
-                    var highScores = await db.QueryAsync<HighScore>($"SELECT h.*, s.id as new_id FROM {highScoreTable} h "
+                    var highScores = await db.QueryAsync<HighScore>($"SELECT h.*, s.id as new_id, s.user_id as new_user_id FROM {highScoreTable} h "
                                                                     + $"LEFT JOIN scores s ON h.score_id = s.legacy_score_id AND s.ruleset_id = {RulesetId} "
                                                                     + "WHERE score_id >= @lastId AND score_id <= @maxProcessableId "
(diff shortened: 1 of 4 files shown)


bdach:
I've expanded an inline comment (https://github.com/ppy/osu-queue-score-statistics/pull/236/commits/6c1e45e5beef217dc36f77a3f32dcdadde2f83a9).

Seems to check out otherwise to me.

bdach (dismissed):

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/ImportHighScoresCommand.cs
```diff
@@ -116,7 +116,7 @@ public async Task<int> OnExecuteAsync(CancellationToken cancellationToken)
                         await checkSlaveLatency(db, cancellationToken);
 
                     Console.WriteLine($"Fetching next scores from {lastProcessedId}...");
-                    var highScores = await db.QueryAsync<HighScore>($"SELECT h.*, s.id as new_id FROM {highScoreTable} h "
+                    var highScores = await db.QueryAsync<HighScore>($"SELECT h.*, s.id as new_id, s.user_id as new_user_id FROM {highScoreTable} h "
```
bdach, line 119: Why is this added to the import command, and why isn't it added to the _watch_ command?

I didn't question this initially, however I noticed some logging looked wonky that I was gonna go and fix (importer not logging deletion numbers), and then I realised that it probably doesn't matter because import is _one-shot_ and what this probably primarily wants to be fixing is the _watcher_, no?

bdach (changes requested):

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Queue/ImportHighScoresCommand.cs
peppy: yes. too many commands.

bdach:
i _think_ this is alright...

the non-high watcher also uses `BatchInserter` but i don't think that's a concern since they insert non-preserved scores to begin with, and i don't think those can go away after pin (because web sets them as preserved, and we're not associating the `scores` rows of non-high with non-high score IDs so there's basically no way to recover anyways)

bdach (approved):

PR #238, opened 2024-03-15
Update user highest rank & rolling rank history

bdach:
- Closes https://github.com/ppy/osu-web/issues/10937

I believe this to be a straight port of the correspondent `osu-web-10` logic, but please do double-check the logic, especially around the 90-day history handling since there's some implicit understanding on how that works that I basically inferred backwards from `osu-web` and `osu-web-10` source.

May be best to deploy after https://github.com/ppy/osu/issues/27174 is resolved.


peppy (approved):
Looks to match expectations

PR #239, opened 2024-03-18
Add command to process total PP for users based on sql

peppy:
Will be used for an initial run to update total PPs to new algorithm.

Previous runs would do batch reprocessing via `osu-performance` using something like:

```bash
./osu-performance sql "select user_id from osu_user_stats where rank_score > 0 and last_played > '2021-01-14' order by rank_score asc" -m osu
```

Of note, see commit message of [ea09969](https://github.com/ppy/osu-queue-score-statistics/pull/239/commits/ea099693b5aa75e50f64e56269caeb81fc032fc1) for reasoning for not updating some pieces.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Performance/PerformanceCommand.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Performance/PerformanceCommand.cs
@@ -69,7 +69,7 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Performance
                     if (userStats == null)
                         return;
 
-                    await TotalProcessor.UpdateUserStatsAsync(userStats, RulesetId, db);
+                    await TotalProcessor.UpdateUserStatsAsync(userStats, RulesetId, db, updateIndex: false);
                     await DatabaseHelper.UpdateUserStatsAsync(userStats, db);
                 }
 
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Performance/UserTotals/UpdateUserTotalsFromSqlCommand.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Performance/UserTotals/UpdateUserTotalsFromSqlCommand.cs
@@ -1,0 +1,41 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using System.ComponentModel.DataAnnotations;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Dapper;
+using JetBrains.Annotations;
+using McMaster.Extensions.CommandLineUtils;
+using osu.Server.QueueProcessor;
+using osu.Server.Queues.ScoreStatisticsProcessor.Helpers;
+
+namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Performance.UserTotals
+{
+    [Command("sql", Description = "Updates the total PP of scoped users.")]
+    public class UpdateUserTotalsFromSqlCommand : PerformanceCommand
+    {
+        [UsedImplicitly]
+        [Required]
+        [Argument(0, Description = "Specify a custom query to limit the scope of users to process.")]
+        public string? Query { get; set; }
+
+        protected override async Task<int> ExecuteAsync(CancellationToken cancellationToken)
+        {
+            using var db = DatabaseAccess.GetConnection();
+
+            LegacyDatabaseHelper.RulesetDatabaseInfo databaseInfo = LegacyDatabaseHelper.GetRulesetSpecifics(RulesetId);
+
+            Console.WriteLine($"Fetching scoped users ({Query})...");
+
+            int[] userIds = (await db.QueryAsync<int>($"SELECT `user_id` FROM {databaseInfo.UserStatsTable} WHERE {Query}")).ToArray();
+
+            Console.WriteLine($"Fetched {userIds.Length} users");
+
+            await ProcessUserTotals(userIds, cancellationToken);
+            return 0;
+        }
+    }
+}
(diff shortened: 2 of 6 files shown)


osu.Server.Queues.ScoreStatisticsProcessor/Commands/Performance/PerformanceCommand.cs
```diff
@@ -69,7 +69,7 @@ protected async Task ProcessUserTotals(int[] userIds, CancellationToken cancella
                     if (userStats == null)
                         return;
 
-                    await TotalProcessor.UpdateUserStatsAsync(userStats, RulesetId, db);
+                    await TotalProcessor.UpdateUserStatsAsync(userStats, RulesetId, db, updateIndex: false);
```
bdach, line 72: If this is going to be purposefully skip updating the rank index, I assume that something is _eventually_ going to ensure that the index is correct?

In other words: there is a scheduled process that updates all users' rank index once a day or something, to ensure there are no three rank 1s or something, correct?

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Performance/UserTotals/UpdateUserTotalsFromSqlCommand.cs
```diff
@@ -0,0 +1,41 @@
...
+
+namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Performance.UserTotals
+{
+    [Command("sql", Description = "Updates the total PP of scoped users.")]
+    public class UpdateUserTotalsFromSqlCommand : PerformanceCommand
+    {
+        [UsedImplicitly]
+        [Required]
+        [Argument(0, Description = "Specify a custom query to limit the scope of users to process.")]
+        public string? Query { get; set; }
```
bdach, line 23: I can't *not* ask... is there a reason for allowing arbitrary sql constructions to be part of the query? Is there nothing more... constrained that I can be offered here? I'm too worried about bobby tables shenanigans to *not* ask this question...

osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserTotalPerformanceProcessor.cs
```diff
@@ -136,21 +139,24 @@ private static async Task updateGlobalRank(UserStats userStats, MySqlConnection
...
-                + $"ON DUPLICATE KEY UPDATE `r{todaysRankColumn}` = @rank",
-                new
-                {
-                    userId = userStats.user_id,
-                    mode = dbInfo.RulesetId,
-                    rank = userStats.rank_score_index
-                }, transaction);
+            // Update user's 90-day rolling rank history for today.
+            // TODO: This doesn't need to be updated here. the osu-web graph always uses `rank_score_index` for the current day.
+            // TODO: code is left for now as it will need to be used in the future for daily processing purposes (once migrated from web-10).
```
bdach, line 144: I know we've had this conversation already, and as stated before I agree that this update doesn't make sense, but I'm kinda not seeing why this pull *specifically* has to disable this if the entire call to `updateGlobalRank()` is if'd out already in the new command...? Which has me worried I might be misunderstanding something about how this is supposed to work.

Either way, don't care about this change as stated, just a bit confused why this came up here at all.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Performance/PerformanceCommand.cs
peppy: Yep, you can check `update_pp_rank_history.php` if curious. I generally pause this while doing any batch updates then let it run the next feasible day.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Performance/UserTotals/UpdateUserTotalsFromSqlCommand.cs
peppy: Reason is flexibility and keeping things simple from a code perspective. See sample query which may and will involve `WHERE`, `JOIN`s, `LIMIT`s and `ORDER`s.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserTotalPerformanceProcessor.cs
peppy: I'd just rather not cause undue database load on what will be a very common query to run (~20-100 times a second for incoming `_high` scores).

I was on the fence of whether to TODO or TODO and comment out. Since you're hesitant I'll switch back to just a TODO for now and we can revisit later.

smoogipoo (dismissed):

osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserTotalPerformanceProcessor.cs
bdach: > I'd just rather not cause undue database load on what will be a very common query to run (~20-100 times a second for incoming `_high` scores).

I figured it was likely going to be this but just wanted to make sure.

Again, I was fine with commenting it out, just mostly wondering why this came up at all :shrug: If you think it's best it's commented out I'd revert to it being commented out, you have a better handle of db perf characteristics than I do.

peppy: nah, let's match web-10 for now and reassesss later.

smoogipoo:
Actually, now I'm confused about this PR. What are you hoping the future to be? All this time we've been moving towards turning off `osu-performance`, but now we're turning off osu-queue-score-statistics instead?

peppy:
2024-03-19 15 06 00@2x

I've highlighted the relevant part. To rephrase in case it wasn't clear, this is switching off **user total PP** processing from `osu-performance` (as mentioned in infra, and using the flag you implemented for this specifically for this purpose in https://github.com/ppy/osu-performance/pull/177).

It is not making any changes to per-score PP. That is something to be tackled separately. The goal here is to stop the flip-flop of user total pp each time you switching clients, which currently can change overall rank and pp by in excess of 30,000 ranks.

The change in this PR turns per-score pp processing **off** for legacy scores, but leaves user total on (see untouched flag at https://github.com/ppy/osu-queue-score-statistics/blob/caea32f739841a63f1feaf3e63217f04cd049b13/osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserTotalPerformanceProcessor.cs#L26)

Of note, legacy scores were not being pushed to the score statistics queue until now, which meant neither of these have been running yet.

smoogipoo (approved):

PR #243, opened 2024-03-25
Delete replays from S3 when cleaning up non-preserved scores

bdach:
One of the pending items in https://github.com/ppy/osu-queue-score-statistics/issues/141.

Tested against a real S3 bucket, seems to work.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/DeleteNonPreservedScoresCommand.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/DeleteNonPreservedScoresCommand.cs
@@ -37,7 +37,7 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Maintenance
                 await deleteCommand.PrepareAsync(cancellationToken);
 
                 var scores = await readConnection.QueryAsync<SoloScore>(new CommandDefinition(
-                    $"SELECT * FROM scores WHERE preserve = 0 AND updated_at < DATE_SUB(NOW(), INTERVAL {preserve_hours} HOUR)", flags: CommandFlags.None, cancellationToken: cancellationToken));
+                    $"SELECT * FROM `scores` WHERE `preserve` = 0 AND `unix_updated_at` < UNIX_TIMESTAMP(DATE_SUB(NOW(), INTERVAL {preserve_hours} HOUR))", flags: CommandFlags.None, cancellationToken: cancellationToken));
 
                 foreach (var score in scores)
                 {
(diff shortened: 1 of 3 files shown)


osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/DeleteNonPreservedScoresCommand.cs
```diff
@@ -25,6 +28,7 @@ public async Task<int> OnExecuteAsync(CancellationToken cancellationToken)
             using (var readConnection = DatabaseAccess.GetConnection())
             using (var deleteConnection = DatabaseAccess.GetConnection())
             using (var deleteCommand = deleteConnection.CreateCommand())
+            using (var s3 = S3.GetClient())
```
bdach, line 31: This will throw on missing `S3_KEY` / `S3_SECRET`. Not sure if there should be a flag on the command to skip s3 for testing or something.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/DeleteNonPreservedScoresCommand.cs
```diff
@@ -33,7 +37,7 @@ public async Task<int> OnExecuteAsync(CancellationToken cancellationToken)
                 await deleteCommand.PrepareAsync(cancellationToken);
 
                 var scores = await readConnection.QueryAsync<SoloScore>(new CommandDefinition(
-                    $"SELECT * FROM scores WHERE preserve = 0 AND updated_at < DATE_SUB(NOW(), INTERVAL {preserve_hours} HOUR)", flags: CommandFlags.None, cancellationToken: cancellationToken));
+                    $"SELECT * FROM `scores` WHERE `preserve` = 0 AND `unix_updated_at` < UNIX_TIMESTAMP(DATE_SUB(NOW(), INTERVAL {preserve_hours} HOUR))", flags: CommandFlags.None, cancellationToken: cancellationToken));
```
bdach, line 40: Please scrutinise this change because the query used the old score structure and as such is currently non-functional against latest master of web.

osu.Server.Queues.ScoreStatisticsProcessor/Helpers/S3.cs
```diff
@@ -0,0 +1,30 @@
...
+using System;
+using Amazon;
+using Amazon.Runtime;
+using Amazon.S3;
+
+namespace osu.Server.Queues.ScoreStatisticsProcessor.Helpers
+{
+    public static class S3
+    {
+        public const string REPLAYS_BUCKET = "score-replays";
```
bdach, line 13: `osu-server-spectator` has this in an envvar. I considered doing the same but given that it was hardcoded in its only usage so far I figured maybe not. Open to changing.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/DeleteNonPreservedScoresCommand.cs
peppy: looks correct 👍

peppy (approved):
Seems fine for a first pass.

bdach:
If there's something more that you expected to see here then I can have a second go.

peppy:
Nope not really. A second pass will happen as I attempt to run this on production over the coming days.

Issue #245, opened 2024-03-27
Add command for applying mod multiplier changes


bdach:
Prerequisite for https://github.com/ppy/osu/issues/27685

PR #246, opened 2024-03-27
Add command for applying simple mod multiplier changes

bdach:
- Closes https://github.com/ppy/osu-queue-score-statistics/issues/245

Opening as draft because while this works, this is the absolute simplest tool for the job at hand (namely, changing classic mod multiplier). Its limitations are:

- Adjusting multipliers on the same scores many times is going to cause precision loss due to rounding.
- This command only works if the multiplier for a mod is not dependent on its settings, ie. it will not work for something like half/double time.

If the properties above are desired _now_, then my proposal is to split this command into two.

- The first will unapply the mod multiplier for _every_ score out there and store that raw value somewhere. I propose that "somewhere" to be the `data` json blob, because it doesn't require costly schema changes to execute.
- The second command will re-apply the multiplier for _every_ score out there by taking raw score from the json blob, deserialising the mods and the total score from it, and using lazer code to retroactively compute the mod multiplier. It'll resemble the verify command but for mod multipliers.


osu.Server.Queues.ScoreStatisticsProcessor/Commands/Score/ChangeModMultiplierCommand.cs
```diff
@@ -0,0 +1,163 @@
...
+            using var conn = DatabaseAccess.GetConnection();
+
+            while (!cancellationToken.IsCancellationRequested)
+            {
+                var scoresToAdjust = (await conn.QueryAsync<ScoreToAdjust>(
+                    "SELECT `id`, `total_score` "
+                    + "FROM `scores` "
+                    + "WHERE `id` BETWEEN @lastId AND (@lastId + @batchSize - 1) "
+                    + "AND `ruleset_id` = @rulesetId "
+                    + "AND JSON_SEARCH(`data`, 'one', @modAcronym, NULL, '$.mods[*].acronym') IS NOT NULL",
```
bdach, line 91: Wasn't sure whether to do this sql-side or c#-side. I *think* this should be *somewhat* painless because `EXPLAIN` pegs this as a where filter over a range PK lookup so it shouldn't be _that_ slow.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Score/ChangeModMultiplierCommand.cs
```diff
@@ -0,0 +1,163 @@
...
+
+        [UsedImplicitly]
+        public async Task<int> OnExecuteAsync(CommandLineApplication app, CancellationToken cancellationToken)
+        {
+            if (RulesetId == null
+                || string.IsNullOrEmpty(ModAcronym)
+                || OldMultiplier == null
+                || NewMultiplier == null)
+            {
+                await Console.Error.WriteLineAsync("One or more required parameters is missing.");
```
peppy, line 55: Kinda weird to have optional arguments which are required (usually I'd use standard arguments in place of this, but I guess your goal is to have things named?)

osu.Server.Queues.ScoreStatisticsProcessor/Commands/ScoreCommands.cs
```diff
@@ -0,0 +1,23 @@
...
+using System.Threading.Tasks;
+using JetBrains.Annotations;
+using McMaster.Extensions.CommandLineUtils;
+using osu.Server.Queues.ScoreStatisticsProcessor.Commands.Score;
+
+namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands
+{
+    [Command(Name = "score", Description = "Runs batch processing on score totals for scores and users.")]
+    [Subcommand(typeof(ChangeModMultiplierCommand))]
+    public class ScoreCommands
```
peppy, line 14: I would have probably just put the command in maintenance, given there's already similar score repair commands in there?

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Score/ChangeModMultiplierCommand.cs
bdach: Yeah I didn't want to have them purely positional to prevent screw-ups from swapping them or something.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/ScoreCommands.cs
bdach: At the time of writing I wasn't sure if there were going to be any others, for e.g. user total recalculations. But I think it's pretty clear to me that it's not required anymore, will move.

bdach: https://github.com/ppy/osu-queue-score-statistics/pull/246/commits/2125c1887481dbf581005d366624e2f0d88f65cb

peppy:
As per discord, not sure why this is close/delete. We will still need this command and the structure likely won't change by much at all. Unless the thought is that we'd run a batch sql (ie. via ASS `mass-operate`) and then run ES as a separate step after we have `total_score_without_mods` and therefore it's redaundant?

PR #248, opened 2024-03-29
Add processor to update last played date in user stats table

bdach:
Not sure how widely used this is exactly, but `osu-web-10` updates this during submission.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Processors/LastPlayedDateProcessor.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Processors/LastPlayedDateProcessor.cs
@@ -1,0 +1,35 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using JetBrains.Annotations;
+using MySqlConnector;
+using osu.Game.Online.API.Requests.Responses;
+using osu.Server.Queues.ScoreStatisticsProcessor.Models;
+
+namespace osu.Server.Queues.ScoreStatisticsProcessor.Processors
+{
+    [UsedImplicitly]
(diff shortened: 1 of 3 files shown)


osu.Server.Queues.ScoreStatisticsProcessor/Processors/LastPlayedDateProcessor.cs
```diff
@@ -0,0 +1,35 @@
...
+        public bool RunOnFailedScores => true;
+        public bool RunOnLegacyScores => false;
+
+        public void RevertFromUserStats(SoloScoreInfo score, UserStats userStats, int previousVersion, MySqlConnection conn, MySqlTransaction transaction)
+        {
+        }
+
+        public void ApplyToUserStats(SoloScoreInfo score, UserStats userStats, MySqlConnection conn, MySqlTransaction transaction)
+        {
+            if (userStats.last_played < score.EndedAt || userStats.playcount == 0)
```
bdach, line 27: This (and the ordering shenanigans) are a bit of a weird complication because the `UserStats` model uses `DateTimeOffset.Now` as the default for `last_played` which is both a little inaccurate if the score takes a while to be processed, and incredibly annoying in testing.

peppy: Even with this commentary I'm still a bit confused. Is there a reason we can't always use `score.EndedAt` and call it a day? Is the worry that scores may be processed out of order?

bdach: > Is the worry that scores may be processed out of order?

Yeah it would be. Think pushing a score for reprocessing out of order for whatever reason to fix a localised issue.

peppy (approved):

PR #251, opened 2024-04-02
Never include self when querying for current rank

peppy:
Floating point inaccuracy could cause an off-by-one if we don't exclude ourselves from the query.

No test because hard to setup a test for this.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserTotalPerformanceProcessor.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserTotalPerformanceProcessor.cs
@@ -84,8 +84,12 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Processors
         private static async Task updateGlobalRank(UserStats userStats, MySqlConnection connection, MySqlTransaction? transaction, LegacyDatabaseHelper.RulesetDatabaseInfo dbInfo)
         {
             // User's current global rank.
-            userStats.rank_score_index = (await connection.QuerySingleAsync<int>($"SELECT COUNT(*) FROM {dbInfo.UserStatsTable} WHERE rank_score > {userStats.rank_score}", transaction: transaction))
-                                         + 1;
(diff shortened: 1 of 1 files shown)


smoogipoo (approved):

peppy:
```sql
> explain SELECT COUNT(*) FROM osu_user_stats WHERE rank_score > (select rank_score from osu_user_stats where user_id = 2);
+----+-------------+----------------+------------+-------+------------------------------+------------+---------+--------+---------+----------+--------------------------+
| id | select_type | table          | partitions | type  | possible_keys                | key        | key_len | ref    | rows    | filtered | Extra                    |
+----+-------------+----------------+------------+-------+------------------------------+------------+---------+--------+---------+----------+--------------------------+
| 1  | PRIMARY     | osu_user_stats | <null>     | range | rank_score,country_acronym_2 | rank_score | 4       | <null> | 1425200 | 100.0    | Using where; Using index |
| 2  | SUBQUERY    | osu_user_stats | <null>     | const | PRIMARY                      | PRIMARY    | 4       | const  | 1       | 100.0    | <null>                   |
+----+-------------+----------------+------------+-------+------------------------------+------------+---------+--------+---------+----------+--------------------------+

2 rows in set
Time: 0.220s

> explain SELECT COUNT(*) FROM osu_user_stats WHERE rank_score > (select rank_score from osu_user_stats where user_id = 2) and user_id != 2;
+----+-------------+----------------+------------+-------+--------------------------------------+------------+---------+--------+---------+----------+--------------------------+
| id | select_type | table          | partitions | type  | possible_keys                        | key        | key_len | ref    | rows    | filtered | Extra                    |
+----+-------------+----------------+------------+-------+--------------------------------------+------------+---------+--------+---------+----------+--------------------------+
| 1  | PRIMARY     | osu_user_stats | <null>     | range | PRIMARY,rank_score,country_acronym_2 | rank_score | 4       | <null> | 1425200 |  50.0    | Using where; Using index |
| 2  | SUBQUERY    | osu_user_stats | <null>     | const | PRIMARY                              | PRIMARY    | 4       | const  | 1       | 100.0    | <null>                   |
+----+-------------+----------------+------------+-------+--------------------------------------+------------+---------+--------+---------+----------+--------------------------+
```

query plan doesn't change, FWIW.

Issue #252, opened 2024-04-10
Some stable scores are missing from leaderboards


peppy:
Reported at https://discord.com/channels/188630481301012481/1097318920991559880/1224976853610270770

![Image](https://github.com/ppy/osu-queue-score-statistics/assets/20418176/0832d394-c0bb-47d7-94e1-64688f92b165)
![Image](https://github.com/ppy/osu-queue-score-statistics/assets/20418176/cd72b3c6-4d07-4774-8f52-27d413634520)

One of the scores in question: https://osu.ppy.sh/scores/osu/4562703492

peppy:
Another weird case of `ranked = 0` on a legacy imported score:

```sql
root@db-master.osu.io:osu> select * from scores where legacy_score_id = 4562703492 and ruleset_id = 0\G;
***************************[ 1. row ]***************************
id                 | 1808993319
user_id            | 7333471
ruleset_id         | 0
beatmap_id         | 3946845
has_replay         | 1
preserve           | 1
ranked             | 0
rank               | S
passed             | 1
accuracy           | 0.990817
max_combo          | 2252
total_score        | 933955
data               | {"mods": [{"acronym": "CL"}], "statistics": {"ok": 15, "great": 1074}, "maximum_statistics": {"great": 1089, "legacy_combo_increase": 1163}}
pp                 | 264.482
legacy_score_id    | 4562703492
legacy_total_score | 72783900
started_at         | <null>
ended_at           | 2024-01-03 16:21:35
unix_updated_at    | 1704298895
build_id           | <null>
```

bdach:
believe this is resolved now.

PR #253, opened 2024-04-10
Add verify step to confirm correct `ranked` flag

peppy:
For reasons I cannot fathom, some imported scores have `ranked` set to zero. These look to be older imports, and new cases are not showing up, so let's just fix those and call it a day.

This assumes that if a score still exists in `osu_scores_high` then it is valid for leaderboards.

Closes https://github.com/ppy/osu-queue-score-statistics/issues/252 (already ran on the affected score; running over all scores overnight).


osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/VerifyImportedScoresCommand.cs
```diff
@@ -188,6 +189,13 @@ public async Task<int> OnExecuteAsync(CancellationToken cancellationToken)
                             sqlBuffer.Append($"UPDATE scores SET pp = {importedScore.HighScore.pp.ToString() ?? "NULL"} WHERE id = {importedScore.id};");
                         }
 
+                        if (!check(importedScore.id, "ranked", importedScore.ranked, true))
```
bdach, line 192: I guess this implicitly assumes that every score in the high tables is automatically ranked, but what about the `hidden` flag in those tables? Are we handling that elsewhere already? I don't think we are, or at least I can't see it happening in code...

peppy: Hidden is exclusively used for restricted users, which is handled as part of the ES indexing step these days. So for the time being we do just import these scores to the new table with no consideration of the flag, and leave that to [es indexer](https://github.com/ppy/osu-elastic-indexer/blob/50f2ab5fe6846d5ad148b6080477fb55ba67ea33/osu.ElasticIndexer/Score.cs#L33) to get right.

bdach (approved):

PR #255, opened 2024-04-11
👷 Trigger remote tag and build on private repo

ThePooN:
@peppy Manual approval for deployments to the `production` environment can be set up in this project's settings, similar to `osu-web`.

### DIFF
--- a/.github/workflows/pack.yml
+++ b/.github/workflows/pack.yml
@@ -70,9 +70,28 @@ jobs:
         uses: actions/checkout@v4
       -
         name: Repository Dispatch
-        uses: peter-evans/repository-dispatch@v3
+        uses: peter-evans/repository-dispatch@ff45666b9427631e3450c54a1bcbee4d9ff4d7c0 # v3.0.0
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/pack.yml
```diff
@@ -70,9 +70,28 @@ jobs:
...
+  trigger_private_build:
+    if: ${{ github.ref_type == 'tag' }}
+    runs-on: ubuntu-latest
+    needs:
+      - push_to_registry
+    environment: production
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v4
```
bdach, line 89: Is this doing anything? It feels unnecessary, if the ref name is in the github context already I don't think this job needs the source for anything anymore?

ThePooN: Yeah, that irks me a little as well. It's in every other project with this kind of workflow in the org and I'm not sure why. Not sure this is necessary to action on since I plan to overhaul all the deployment processes to get rid of most plumbing on the private side.

peppy (approved):

PR #257, opened 2024-04-14
Upgrade game packages

peppy:
Pulls in https://github.com/ppy/osu/pull/27729#event-12243877951, importantly.


peppy:
https://github.com/ppy/osu-queue-processor/pull/30

bdach (approved):

PR #258, opened 2024-04-15
Check ranked state when retrieving user best scores

bdach:
`DatabaseHelper.GetUserBestScoreFor()` did not check the value of `ranked` for the rows it was looking through to find the user's best score, which means that it could wrongly select an unranked score (i.e. one set back when a map was in qualified).

This potentially impacts: user total ranked score and user total letter rank counts in all rulesets (including mania keymodes). Please see added test cases for explanation as to how. The worst case scenario is that the user's real best score would not be counted towards those stats because of an unranked score with a higher total.

I don't believe there to be a sane path for retroactively fixing that (if it ever even happened, which I don't know - I only realised this could be a possibility when working on medals / https://github.com/ppy/osu-queue-score-statistics/pull/254). Potentially reprocessing those scores could work, if they can be located easily - which is going to be the bigger problem here.


smoogipoo (approved):
At face value this looks correct.

peppy (approved):

PR #263, opened 2024-04-26
Fix missing beatmap when running score PP processing via CLI

smoogipoo:
Will fix the missing PP sheets seen in:
https://github.com/ppy/osu/pull/28005#issuecomment-2079283702
https://github.com/ppy/osu/pull/27303#issuecomment-2072558859


peppy (approved):

Jenn2188 (changes requested):
Fuck you

PR #198, opened 2024-01-11
Update to use new table structure

peppy:
Running a test import currently, but still appreciate review on this.

Sorry it's a big one..

PR #264, opened 2024-05-01
Decouple score processor from `SoloScoreInfo` model

bdach:
RFC. PRing early as draft to not invest more time into this rabbit-hole before I get buy-in. Nerd essay incoming, please bear with me as I attempt to explain myself.

Where do I even begin... This is a result of several layers of yak-shaving.

- Yak level 0: I want to start on https://github.com/ppy/osu/issues/16480. I go to the client code and I see the absolute mess that is the results screen data structure juggling across different paths (multiplayer/playlists/solo/local scores) with different things using different models and relying on things being there that aren't necessarily always there, and I decide that the results screen hierarchy should use `IScoreInfo`.
- Yak level 1: I look at what it takes to move things to `IScoreInfo`. The answer is: a lot, but a low-hanging fruit was to add `(Maximum)Statistics` to that interface, because it exists already on both implementations (`ScoreInfo` and `SoloScoreInfo`). Things go relatively smoothly until I notice [these two helpers](https://github.com/ppy/osu/blob/61350ebf6ea89605997d91a71e9de14540aad9cd/osu.Game/Scoring/Legacy/ScoreInfoExtensions.cs#L17-L21).
 
  It would seem that these can use `IScoreInfo` now, but knowing that the second method is used server-side in this repository, I go to check server code, and come to the horrifying realisation that they cannot, because on `SoloScoreInfo` in this repository it can actually hold that `soloScoreInfo.RulesetID != soloScoreInfo.Ruleset.OnlineID`, because the former is the _actual_ ID of the ruleset that the score was set on, and the second [actually silently will go through the beatmap](https://github.com/ppy/osu/blob/61350ebf6ea89605997d91a71e9de14540aad9cd/osu.Game/Online/API/Requests/Responses/SoloScoreInfo.cs#L152) and [server-side we set the beatmap to this](https://github.com/ppy/osu-queue-score-statistics/blob/b3887b3456de2f04b362f5ae822f39e8e0ad4b97/osu.Server.Queues.ScoreStatisticsProcessor/ScoreStatisticsQueueProcessor.cs#L180-L183), which is actually the raw database row underneath and will have the ruleset ID _before_ conversion. I decide that this is completely pants on head bonkers and decide to change it.
- Yak level 2: We are here.

Generally I find that the `SoloScoreInfo` convertery probably made more sense before https://github.com/ppy/osu-queue-score-statistics/pull/198, as it made sense to just put the json where the json need to go verbatim in the same format as client deserialised it, but now it's pretty much not that helpful in that respect and makes it actively way too difficult to make any changes client-side as you have to remember that Thing X is used in Place Y and if you forget about it you silently break things whenever the next game package bump happens and now you have a whole new mystery to unravel.

I find myself leaning towards this diff being a Good Thing but I want to get second opinions before I commit too hard. Note that this will also mean that I will have to update the already-merged hush-hush medal stuff to update to this (which is probably going to be _fun_ to both do _and_ review), so this is _not_ all of the changes required yet.

Probably the actual best tangible argument in favour of this change is that it makes https://github.com/ppy/osu-queue-score-statistics/commit/121f3023f114c83522dd949ead8aa4d1227db0c3 possible. https://github.com/ppy/osu-queue-score-statistics/commit/b03b4f35a309089cf572e16df856228466b9a76c is also nice but I believe it could be applied on `master` already in an altered form, I just noticed that it was stupid while doing the rest of this.

This probably needs more testing / scrutiny in the weird parts that still have to interface with game code (namely `SoloScore.ToScoreInfo()` and the implementation of `IBeatmapOnlineInfo` / `GetLegacyBeatmapConversionDifficultyInfo()` in `Beatmap`). I only checked that this passes tests. Again, before I commit more time into this, I want at least an acknowledgement that we do want to go this way.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor.Tests/MedalAwarderTest.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor.Tests/MedalAwarderTest.cs
@@ -83,7 +83,7 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Tests
             // Usually osu-web would do this.
             using (var db = Processor.GetDatabaseConnection())
             {
-                db.Execute($"INSERT INTO osu_user_achievements (achievement_id, user_id, beatmap_id) VALUES ({awarded.Medal.achievement_id}, {awarded.Score.UserID}, {awarded.Score.BeatmapID})");
+                db.Execute($"INSERT INTO osu_user_achievements (achievement_id, user_id, beatmap_id) VALUES ({awarded.Medal.achievement_id}, {awarded.Score.user_id}, {awarded.Score.beatmap_id})");
             }
         }
 
--- a/osu.Server.Queues.ScoreStatisticsProcessor.Tests/PerformanceProcessorTests.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor.Tests/PerformanceProcessorTests.cs
@@ -9,7 +9,6 @@ using MySqlConnector;
 using osu.Framework.Extensions.TypeExtensions;
 using osu.Framework.Localisation;
 using osu.Game.Online.API;
-using osu.Game.Online.API.Requests.Responses;
 using osu.Game.Rulesets.Catch.Mods;
 using osu.Game.Rulesets.Mania.Mods;
 using osu.Game.Rulesets.Mods;
@@ -144,7 +143,7 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Tests
             };
 
             foreach (var mod in mods)
-                Assert.True(ScorePerformanceProcessor.AllModsValidForPerformance(new SoloScoreInfo(), new[] { mod }), mod.GetType().ReadableName());
+                Assert.True(ScorePerformanceProcessor.AllModsValidForPerformance(new SoloScore(), new[] { mod }), mod.GetType().ReadableName());
         }
 
         [Fact]
@@ -176,7 +175,7 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Tests
             };
 
             foreach (var mod in mods)
-                Assert.False(ScorePerformanceProcessor.AllModsValidForPerformance(new SoloScoreInfo(), new[] { mod }), mod.GetType().ReadableName());
+                Assert.False(ScorePerformanceProcessor.AllModsValidForPerformance(new SoloScore(), new[] { mod }), mod.GetType().ReadableName());
         }
 
         [Fact]
@@ -204,7 +203,7 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Tests
             };
 
             foreach (var mod in mods)
-                Assert.False(ScorePerformanceProcessor.AllModsValidForPerformance(new SoloScoreInfo(), new[] { mod }), mod.GetType().ReadableName());
+                Assert.False(ScorePerformanceProcessor.AllModsValidForPerformance(new SoloScore(), new[] { mod }), mod.GetType().ReadableName());
         }
 
         [Fact]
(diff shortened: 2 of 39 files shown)


peppy (approved):
I have no issues with this direction.

bdach:
I've done some further testing which yielded the above commits. However I'm not yet done so not undrafting quite yet.

Of note are: https://github.com/ppy/osu-queue-score-statistics/pull/264/commits/5699222e2ed3f2bea33c7f0ed202883397cf69e3, and https://github.com/ppy/osu-queue-score-statistics/pull/264/commits/6bc4abe7f75c7c71951c38bf490c69ee0472904d. That second one is especially egregious because it is silent breakage due to all of the mazey conversions between various score-shaped fiascos and I only caught it by manually inspecting _every single usage_ of the model inside the difficulty calculators. When I found that taiko convert check I expected it to blow up on a null but no, instead it [just silently reads the wrong thing because of this!](https://github.com/ppy/osu-queue-score-statistics/pull/264/commits/6bc4abe7f75c7c71951c38bf490c69ee0472904d)

Silent fallbacks are such a mistake. I will attempt to be a _lot_ more annoying about clean models from this point on.

I still want to do one more visual pass over the diff, and then run sheets for pp and score for _all_ rulesets. So this will probably be stuck in draft for like another day until my box crunches all that.

---

If you wish to follow along with the madness here's a log of things that I was checking:

<details>
<summary>beware here be dragons</summary>

- [x] `ToScoreInfo()`
	- [x] passed to `GetDisplayScore()` extension in multiple processors.
	      `GetDisplayScore()` reads:
			- [x] `ScoreInfo.Ruleset.OnlineID`
			- [x] `ScoreInfo.TotalScore`
			- [x] `ScoreInfo.MaximumStatistics`
	- [x] passed to `PerformanceCalculator.Calculate()` in `ScorePerformanceProcessor`
		- `CatchPerformanceCalculator.Calculate()` reads:
			- [x] `ScoreInfo.GetCount{300,100,50,Katu,Miss}()` -> `ScoreInfo.Statistics`
			- [x] `ScoreInfo.Mods`
		- `ManiaPerformanceCalculator.Calculate()` reads:
			- [x] `ScoreInfo.Statistics`
			- [x] `ScoreInfo.Mods`
		- `OsuPerformanceCalculator.Calculate()` reads:
			- [x] `ScoreInfo.Accuracy`
			- [x] `ScoreInfo.MaxCombo`
			- [x] `ScoreInfo.Statistics`
			- [x] `ScoreInfo.Mods`
		- `TaikoPerformanceCalculator.Calculate()` reads:
			- [x] `ScoreInfo.Statistics`
			- [x] `ScoreInfo.BeatmapInfo!.Ruleset.OnlineID`
			- [x] `ScoreInfo.Mods`
- [x] `Beatmap : IBeatmapOnlineInfo`
	- [x] used for `DifficultyAttributes.FromDatabaseAttributes()` in `BeatmapStore`
		- `DifficultyAttributes.FromDifficultyAttributes()` reads nothing from the `onlineInfo`
		- `CatchDifficultyAttributes.FromDifficultyAttributes()` reads nothing from the `onlineInfo`
		- `ManiaDifficultyAttributes.FromDifficultyAttributes()` reads nothing from the `onlineInfo`
		- `OsuDifficultyAttributes.FromDifficultyAttributes()` reads:
			- [x] `DrainRate`
			- [x] `CircleCount`
			- [x] `SliderCount`
			- [x] `SpinnerCount`
		- `TaikoDifficultyAttrobutes.FromDifficultyAttributes()` reads nothing from the `onlineInfo`
- [x] `Beatmap.GetLegacyConversionDifficultyInfo()`
	- [x] passed to `StandardisedScoreMigrationTools.UpdateFromLegacy()` in `BatchInserter`
		- then passed to `ILegacyScoreSimulator.GetLegacyScoreMultiplier()`
			- `CatchLegacyScoreSimulator` reads nothing from there
			- `ManiaLegacyScoreSimulator` reads:
				- [x] `SourceRuleset`
				- [x] `DrainRate`
				- [x] `CircleSize`
				- [x] `OverallDifficulty`
				- [x] `ApproachRate`
				- [x] `TotalObjectCount`
				- [x] `EndTimeObjectCount`
			- `OsuLegacyScoreSimulator` reads nothing from there
			- `TaikoLegacyScoreSimulator` reads nothing from there

- [x] do manual audit of the above
- [x] for the display score stuff maybe do draft unit tests that exercises the other rulesets and compare result against `master`
- [ ] for the rest, do a diffcalc run, diffing the two queue processor versions





</details>

bdach:
Sheets:

- taiko: https://docs.google.com/spreadsheets/d/11TnhGI7eP8LvTswepRIfEDCRe9_n3G6u6iUlxpP7eBY/edit#gid=1348981755
- catch: https://docs.google.com/spreadsheets/d/1f1JFx2apqQOvDpnkBfFziuybOSCLXBQIZmgh3HQ28d0/edit#gid=440333105
- osu!: https://docs.google.com/spreadsheets/d/1iPAzhaLrZuSUr48hYE73MPibSizQ0n39igOR3dvY0Fg/edit#gid=293445165
- mania: https://docs.google.com/spreadsheets/d/1bzXHpKovsgenEbP2eDwbsu0t7AopT8UUvY4O96RJ3_c/edit#gid=702742875

(If you're wondering why the titles of the spreadsheet have 2 different hashes: I ran the sheets with

```
OSU_A=https://github.com/bdach/osu/tree/master
OSU_B=https://github.com/bdach/osu/tree/master-2
SCORE_PROCESSOR_A=https://github.com/ppy/osu-queue-score-statistics
SCORE_PROCESSOR_B=https://github.com/ppy/osu-queue-score-statistics/pull/264
```

`OSU_B` has a single empty commit on top of `OSU_A`. This is to fool the sheet machinery into not recognising that the two commits are the same. The sheet generation relies heavily on commit SHA, by way of naming databases for which it's running diffcalc using the SHAs, so if I had just ran it on master both times, it would render the whole exercise essentially pointless.)

Empty as I'd hoped. I don't think there's anything else I can think of to test, so undrafting.

smoogipoo (approved):

PR #266, opened 2024-05-13
Improve error handling

peppy:
Just some reactionary changes based on recent exceptions on production

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Helpers/UserTotalPerformanceAggregateHelper.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Helpers/UserTotalPerformanceAggregateHelper.cs
@@ -44,6 +44,12 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Helpers
                 totalAccuracy *= 100.0 / (20 * (1 - Math.Pow(0.95, groupedScores.Length)));
             }
 
+            if (double.IsNegative(totalPp) || double.IsNaN(totalPp) || double.IsInfinity(totalPp))
+                throw new InvalidOperationException($"Calculating total PP for user_id:{scores.First().user_id} resulted in invalid value ({totalPp}");
+
+            if (double.IsNegative(totalAccuracy) || double.IsNaN(totalAccuracy) || double.IsInfinity(totalAccuracy))
+                throw new InvalidOperationException($"Calculating total accuracy for user_id:{scores.First().user_id} resulted in invalid value ({totalAccuracy}");
+
             return ((float)totalPp, (float)totalAccuracy);
         }
     }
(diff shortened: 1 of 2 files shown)


osu.Server.Queues.ScoreStatisticsProcessor/Helpers/UserTotalPerformanceAggregateHelper.cs
```diff
@@ -44,6 +44,12 @@ public static (float totalPp, float totalAccuracy) CalculateUserTotalPerformance
                 totalAccuracy *= 100.0 / (20 * (1 - Math.Pow(0.95, groupedScores.Length)));
             }
 
+            if (double.IsNegative(totalPp) || double.IsNaN(totalPp) || double.IsInfinity(totalPp))
+                throw new InvalidOperationException($"Calculating total PP for user_id:{scores.First().user_id} resulted in invalid value ({totalPp}");
+
+            if (double.IsNegative(totalAccuracy) || double.IsNaN(totalAccuracy) || double.IsInfinity(totalAccuracy))
+                throw new InvalidOperationException($"Calculating total accuracy for user_id:{scores.First().user_id} resulted in invalid value ({totalAccuracy}");
```
bdach, line 51: Could probably check if this is in [0;100] range here if we're doing this?

osu.Server.Queues.ScoreStatisticsProcessor/ScoreStatisticsQueueProcessor.cs
```diff
@@ -160,6 +161,15 @@ protected override void ProcessResult(ScoreItem item)
...
             try
             {
+                SentrySdk.ConfigureScope(scope =>
+                {
+                    scope.SetTag("score_id", item.Score.id.ToString());
+                    scope.User = new SentryUser
+                    {
+                        Id = item.Score.user_id.ToString(),
+                    };
+                });
```
bdach, line 171: I guess this theoretically works as long as the processor is single-threaded but I'm kinda skeeved out by this as the possibility of mislabeling events exists. Could possibly use [something like this](https://github.com/getsentry/sentry-docs/issues/5460#issuecomment-1234516412) instead, dunno.

peppy: I'm not sure there's a guarantee of single-threadedness.

Just naively assumed sentry would scope per thread but maybe that's not the case. WIll refactor slightly.

peppy: Note sure `PushScope` is enough because it seems we're running in the default `GlobalMode` which doesn't seems to work well with it, but we'll see.

bdach: I've tested and the `PushScope()` stuff seems to cause the tags to not be pushed, so force-pushing out.

PR #268, opened 2024-05-21
Populate score without mods during ongoing imports & add command for backpopulation of it on existing scores

bdach:
- End goal is https://github.com/ppy/osu/issues/27685.
- [x] Depends on https://github.com/ppy/osu-web/pull/11211
- [x] Depends on https://github.com/ppy/osu/pull/28277
- [x] Submission should be turned off on old clients that don't send total score without mods before running the command here

This PR ensures that the legacy score watchers populate `total_score_without_mods` correctly to the json (to match https://github.com/ppy/osu-web/pull/11211), and also features a command to fill in that data wherever it's missing, using an approximate estimate.

Only tested locally on a small `scores` table, so buyer beware.


peppy (approved):

peppy:
Just a self heads-up that this still needs to be run at some point, with consideration given to table size expansion when it is.

PR #269, opened 2024-05-21
Add command for recalculating mod multipliers

### LINKED ISSUE #245: Add command for applying mod multiplier changes

bdach:
- Closes https://github.com/ppy/osu-queue-score-statistics/issues/245.
- [x] Depends on https://github.com/ppy/osu-queue-score-statistics/pull/268 for code.
- [ ] Depends on game NuGet packages which include the new score multipliers
- [ ] The command from the above pull should additionally be ran before trying this. When trying to run the command added here, all scores should have `total_score_without_mods` in the json.


bdach:
I'll look into the test failures, but a thought that has just occurred to me is that this PR might be going too far.

By which I mean we may want to skip recalculating mod multipliers on leaderboards with significance - I'm thinking multiplayer, playlists, and daily challenge. Which is to say, any multiplayer score would be excluded from having its multiplier recalculated (but also would need to become unranked in the process, which also has pp implications).

Thoughts?

tsunyoku:
@peppy could I ask that this change is looked at some point soon? With the recent creation of https://discord.com/channels/546120878908506119/1373971097434980463 and conversation happening there, I'm at a point where I'm not sure how exactly we're supposed to go ahead publicly if there's no framework to actually facilitate any changes.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/RecalculateModMultipliersCommand.cs
```diff
@@ -0,0 +1,147 @@
...
+                {
+                    score.beatmap = beatmapsById[score.beatmap_id];
+                    var scoreInfo = score.ToScoreInfo();
+
+                    if (scoreInfo.TotalScoreWithoutMods == 0 && scoreInfo.TotalScore != 0)
+                    {
+                        throw new InvalidOperationException($"Score with ID {score.id} has {scoreInfo.TotalScore} total score but {scoreInfo.TotalScoreWithoutMods} total score without mods. "
+                                                            + $"This is likely to indicate that {nameof(scoreInfo.TotalScoreWithoutMods)} was not correctly backpopulated on all scores "
+                                                            + "(or there is a process pushing new scores that was not updated to populate the field).");
+                    }
```
bdach, line 89: This will likely need further adjusting:

- Will die on missing beatmap in ways similar to what https://github.com/ppy/osu-queue-score-statistics/pull/371 was fixing
- The hard throw on zero total score without mods but nonzero total score will likely need to be changed to a soft skip or something. @peppy not sure what sort of visibility you want on skips here (just log count of skipped rows?)

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/RecalculateModMultipliersCommand.cs
```diff
@@ -0,0 +1,147 @@
...
+
+                    double multiplier = 1;
+
+                    foreach (var mod in scoreInfo.Mods)
+                        multiplier *= mod.ScoreMultiplier;
```
bdach, line 94: Depends on client changes (the new multipliers actually being present in client source), and will require upgrading in line with the new API

bdach (commented):
Just noting current known deficiencies / blockers after 2 years of not seeing this diff.

Will want to re-run with the new multiplier API at least locally and re-check that this is performing within expected parameters.

peppy (approved):

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/RecalculateModMultipliersCommand.cs
peppy: To recap, we're currently populating the majority of scores, but there could be some which don't get the population due to deleted/missing beatmaps or other edge case issues.

I think skipping here with log output is fine for such scores.

PR #270, opened 2024-05-22
Add partition cache for `user_rank_index` updates

peppy:
Similar to what is done in web-10.

- [x] Add a few more tests which better test new cache (current coverage is only for a single entry in `user_stats`)

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserTotalPerformanceProcessor.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserTotalPerformanceProcessor.cs
@@ -25,6 +25,10 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Processors
 
         public bool RunOnLegacyScores => true;
 
+        // [ruleset_id, [rank_score, count_users_above]]
+        private readonly ConcurrentDictionary<int, ConcurrentDictionary<int, int>> rankScoreIndexPartitionCache =
+            new ConcurrentDictionary<int, ConcurrentDictionary<int, int>>();
+
         public void RevertFromUserStats(SoloScore score, UserStats userStats, int previousVersion, MySqlConnection conn, MySqlTransaction transaction)
         {
         }
@@ -82,7 +86,7 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Processors
                 await updateGlobalRank(userStats, connection, transaction, dbInfo);
         }
 
-        private static async Task updateGlobalRank(UserStats userStats, MySqlConnection connection, MySqlTransaction? transaction, LegacyDatabaseHelper.RulesetDatabaseInfo dbInfo)
+        private async Task updateGlobalRank(UserStats userStats, MySqlConnection connection, MySqlTransaction? transaction, LegacyDatabaseHelper.RulesetDatabaseInfo dbInfo)
         {
             // User's current global rank.
             userStats.rank_score_index = await getUserRankScoreIndex(userStats, connection, transaction, dbInfo);
@@ -126,13 +130,9 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Processors
                 }, transaction);
         }
 
-        // [ruleset_id, [rank_score, count_users_above]]
-        private static readonly ConcurrentDictionary<int, ConcurrentDictionary<int, int>> rank_score_index_partition_cache =
-            new ConcurrentDictionary<int, ConcurrentDictionary<int, int>>();
-
-        private static async Task<int> getUserRankScoreIndex(UserStats userStats, MySqlConnection connection, MySqlTransaction? transaction, LegacyDatabaseHelper.RulesetDatabaseInfo dbInfo)
+        private async Task<int> getUserRankScoreIndex(UserStats userStats, MySqlConnection connection, MySqlTransaction? transaction, LegacyDatabaseHelper.RulesetDatabaseInfo dbInfo)
         {
-            var rulesetCache = rank_score_index_partition_cache.GetOrAdd(dbInfo.RulesetId, _ => new ConcurrentDictionary<int, int>());
+            var rulesetCache = rankScoreIndexPartitionCache.GetOrAdd(dbInfo.RulesetId, _ => new ConcurrentDictionary<int, int>());
 
             const int partition_size = 100;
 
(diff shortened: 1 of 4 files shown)


bdach:
Would something like the following work for testing?

```diff
diff --git a/osu.Server.Queues.ScoreStatisticsProcessor.Tests/PerformanceProcessorTests.cs b/osu.Server.Queues.ScoreStatisticsProcessor.Tests/PerformanceProcessorTests.cs
index f8d76be..5d9ac80 100644
--- a/osu.Server.Queues.ScoreStatisticsProcessor.Tests/PerformanceProcessorTests.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor.Tests/PerformanceProcessorTests.cs
@@ -448,6 +448,67 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Tests
             });
         }
 
+        [Fact]
+        public async Task PartitionCaching()
+        {
+            // simulate fake users to beat as we climb up ranks.
+            // this is going to be a bit of a chonker query...
+            using var db = Processor.GetDatabaseConnection();
+
+            var stringBuilder = new StringBuilder();
+
+            stringBuilder.Append("INSERT INTO osu_user_stats (`user_id`, `rank_score`, `rank_score_index`, "
+                                 + "`accuracy_total`, `accuracy_count`, `accuracy`, `accuracy_new`, `playcount`, `ranked_score`, `total_score`, "
+                                 + "`x_rank_count`, `xh_rank_count`, `s_rank_count`, `sh_rank_count`, `a_rank_count`, `rank`, `level`) VALUES ");
+
+            for (int i = 0; i < 1000; ++i)
+            {
+                if (i > 0)
+                    stringBuilder.Append(',');
+
+                stringBuilder.Append($"({1000 + i}, {25 * (1000 - i)}, {i + 1}, 0, 0, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1)");
+            }
+
+            await db.ExecuteAsync(stringBuilder.ToString());
+
+            var firstBeatmap = AddBeatmap(b => b.beatmap_id = 1);
+
+            SetScoreForBeatmap(firstBeatmap.beatmap_id, s =>
+            {
+                s.Score.preserve = s.Score.ranked = true;
+                s.Score.pp = 150; // ~152 pp total, including bonus pp
+            });
+
+            WaitForDatabaseState("SELECT `rank_score_index` FROM `osu_user_stats` WHERE `user_id` = @userId", 995, CancellationToken, new
+            {
+                userId = 2,
+            });
+
+            SetScoreForBeatmap(firstBeatmap.beatmap_id, s =>
+            {
+                s.Score.preserve = s.Score.ranked = true;
+                s.Score.pp = 180; // ~184 pp total, including bonus pp
+            });
+
+            WaitForDatabaseState("SELECT `rank_score_index` FROM `osu_user_stats` WHERE `user_id` = @userId", 994, CancellationToken, new
+            {
+                userId = 2,
+            });
+
+            var secondBeatmap = AddBeatmap(b => b.beatmap_id = 2);
+
+            SetScoreForBeatmap(secondBeatmap.beatmap_id, s =>
+            {
+                s.Score.preserve = s.Score.ranked = true;
+                s.Score.pp = 300; // ~486 pp total, including bonus pp
+            });
+
+            WaitForDatabaseState("SELECT `rank_score_index` FROM `osu_user_stats` WHERE `user_id` = @userId", 982, CancellationToken, new
+            {
+                userId = 2,
+            });
+        }
+
         [Fact]
         public async Task UserDailyRankUpdates()
         {

```

osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserTotalPerformanceProcessor.cs
```diff
@@ -128,5 +125,38 @@ private static async Task updateGlobalRank(UserStats userStats, MySqlConnection
...
+            if (RNG.NextSingle() > 0.999) rulesetCache.Remove(partitionCutoff, out int _);
+
+            // This query is indexed as much as it can be, but for COUNT(*) across millions of rows it can still be slow.
+            // Cache the majority lookup by partitioning over `pp` reduces the per-query lookup overhead.
+            int partitionCount = rulesetCache.GetOrAdd(partitionCutoff, cutoff => connection.QuerySingle<int>(
+                $"SELECT COUNT(*) FROM {dbInfo.UserStatsTable} WHERE rank_score > @cutoff",
+                new { cutoff }, transaction: transaction));
+
+            int remainingCount = await connection.QuerySingleAsync<int>(
+                $"SELECT COUNT(*) FROM {dbInfo.UserStatsTable} WHERE rank_score BETWEEN @rankScoreCutoff AND @partitionCutoff AND user_id != @userId",
```
bdach, line 151: Bit of a nitpick but `BETWEEN` is both-ends inclusive. Previous query was `rank_score > @rankScoreCutoff` and now it's a union of `rank_score > @partitionCutoff` and `rank_score >= @rankScoreCutoff AND rank_score <= @partitionCutoff`, netting `rank_score >= @rankScoreCutoff` in general, so it's a slight behavioural change. In practice likely won't matter much because floats but it might matter for something like zero? dunno.

If we wanted to be super proper you'd want to have

```suggestion
                $"SELECT COUNT(*) FROM {dbInfo.UserStatsTable} WHERE rank_score > @rankScoreCutoff AND rank_score <= @partitionCutoff AND user_id != @userId",
```

I believe.

bdach (commented):
Seems to pass the eye test. I dunno about the 0.1% random invalidation thing but I assume this sort of thing was battle-tested elsewhere so I'm not going to pretend I know more.

With partitioning in increments of 100 this should net ~270 partitions in production, which seems more than safe memory footprint wise.

peppy:
> Seems to pass the eye test. I dunno about the 0.1% random invalidation thing but I assume this sort of thing was battle-tested elsewhere so I'm not going to pretend I know more.

Not really battle-tested. Just something we'd tweak as required. The hope is a cache hit rate of 99.9% will alleviate enough load while also not reducing accuracy of calculations. A time-based expiry rule might probably be more correct but a bit more effort to implement, manage and validate.

peppy:
> Not sure if you care but cache hit rate metrics can now be easily reported to ddog for inspection, MemoryCache provides this.

May add as a follow up but i'd mostly hope this is hands-off. If anything we may want to adjust the partition size downwards, but we'll see.

PR #271, opened 2024-05-30
Reduce verbosity of `verify-imported-scores` command

smoogipoo:
The cursor trick doesn't work on docker/etc, making this command very spammy: https://github.com/ppy/osu/actions/runs/9298230620/job/25589772915

In this case, I'll set the `--quiet` flag and live with the single line of output at the end.


bdach (approved):

peppy (approved):
🤷

PR #274, opened 2024-07-15
Enforce real time difficulty for mod combinations which are not stored in the database

tsunyoku:
Old discussion about the idea available on Discord: https://discord.com/channels/188630481301012481/380598781432823815/1245654906761777267.

Talked with smoogi about the idea and seems like a good way to get the ball rolling. This will allow for awarding pp on rates for double time other than the default (1.5x) and/or awarding pp for non-legacy mods which have difficulty changes. There's technically some edge cases here when a mod has settings which won't effect difficulty changes and will still go through realtime, but I don't think this is a major issue (and doesn't have a pretty solution). Classic is excluded from the "non-legacy mods" because default CL settings is treated the same as non-CL for now - this might become messy once difficulty calculation changes that effect non-CL specifically are deployed.

I've renamed the `REALTIME_DIFFICULTY` env var to `ALWAYS_REALTIME_DIFFICULTY` to try and reduce confusion over *when* realtime calculations will be used. When merging this'll need to be accounted for to avoid accidentally enabling realtime difficulty *always*.

To help smoogi/the team with profiling and the implications of realtime calculations on production, I've added a `calculate-realtime-difficulty-attributes` datadog timer with tags for the mods and ruleset. It's worth noting that the time will include the `.osu` file download time. This should help figuring out if realtime is too bad to use always, since long-term that would be the better option.

`isRankedLegacyMod` exists as an easy solution to figure out if the mod is going to be cached in the database. I don't really like this method in general but I don't know of a better way to do it...

There's an added test using `OsuModMuted` and `OsuModTraceable` as these are the only 2 mods which are non-legacy, so would be expected to go down the realtime path.


bdach:
> Talked with smoogi about the idea

Was this in DMs or something? I'd rather not proliferate that form of communication.

I don't disagree with the decision to do it, but I'd rather not play telephone game over such matters.

> It's worth noting that the time will include the .osu file download time

I would hope that production would not download beatmaps. Or at least hope that it caches them if it *has* to download.

> I wanted to write tests for these cases, however with this diff alone it's not really possible as in all cases `mustUseRealtimeDifficulty` would be `true` the mod will have `Ranked` as `false` so it'll never reach this block. Tests can be added in the future when this is merged.

I don't understand. How does merging this help write tests? Either tests can be written for this now *and* after merge, *or* neither now or after merge *until* something else happens?

bdach (changes requested):

smoogipoo:
> Was this in DMs or something? I'd rather not proliferate that form of communication.

Only briefly. It was not discussed in depth but I said it's a good path to go down.

osu.Server.Queues.ScoreStatisticsProcessor/Stores/BeatmapStore.cs
```diff
@@ -107,12 +127,33 @@ public static async Task<BeatmapStore> CreateAsync(MySqlConnection connection, M
...
+                or ModPerfect
+                or ModSuddenDeath
+                or ModNightcore
+                or ModDoubleTime
+                or ModRelax
+                or ModHalfTime
+                or ModFlashlight
+                or ModCinema
+                or ModAutoplay
+                or ModScoreV2;
```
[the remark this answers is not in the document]
tsunyoku: Yeah I think I forgot about non-standard. Your comment about wanting to enumerate "ranked mods from stable" is accurate but I can't figure out a better way than this listing of mods. I'll add the non-standard mods.

bdach:
> It would _allow_ for setting `Ranked` on a mod that relies on the realtime path. Merging won't immediately make tests writable, I'm just highlighting that reaching this code with the current set of ranked mods is impossible.

This makes much more sense as an explanation.

> This is existing logic, the download path is settable with an env var so I'm not sure if there's some internal service communication that has some caching or something but if not then I'm open to replacement suggestions but I don't know how the infra works at that level.

Sure, it is, it has also never been ran on production, which means that figuring out whether or not the live version should download beatmaps or not is a valid review concern in my eyes.

osu.Server.Queues.ScoreStatisticsProcessor/Stores/BeatmapStore.cs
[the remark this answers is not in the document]
bdach: I'd also run a double-check as to whether there are any stable mods that are only ranked on certain rulesets, which might break this method further.

tsunyoku:
> figuring out whether or not the live version should download beatmaps or not is a valid review concern in my eyes.
Yeah I 100% agree, just not sure if it's something I can personally weigh in on. I don't know enough about the infrastructure to comment.

osu.Server.Queues.ScoreStatisticsProcessor/Stores/BeatmapStore.cs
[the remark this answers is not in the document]
tsunyoku: Did a double-check - this seems to only apply for `ManiaModHardRock` (which I've now fixed)

tsunyoku:
> It would allow for setting Ranked on a mod that relies on the realtime path. Merging won't immediately make tests writable, I'm just highlighting that reaching this code with the current set of ranked mods is impossible.

I realise this isn't *technically* true because mods like muted and traceable are ranked which do not affect difficulty calculations, but fit under the conditions for going through realtime calculations. Should I add tests for those cases? Is it a concern that these mods are going through realtime calculations when they technically have no effect on difficulty (*yet*)?

bdach:
> Should I add tests for those cases?

I'll always take tests if it is possible to write them.

tsunyoku:
Sorry about all the line ending commits.. I have something misconfigured somewhere and it keeps converting line endings :(

peppy:
I've cleaned up commit history with the line ending shenanigans. Please fix your settings before committing again.

README.md
```diff
@@ -119,9 +119,9 @@ Whether to process user total stats. Set to `0` to disable processing.
 
 Default is unset (processing enabled).
 
-### REALTIME_DIFFICULTY
+### ALWAYS_REALTIME_DIFFICULTY
```
peppy, line 122: How about we call this `PREFER_REALTIME_DIFFICULTY`? I think it makes better grammatical sense.

osu.Server.Queues.ScoreStatisticsProcessor/Stores/BeatmapStore.cs
```diff
@@ -79,7 +93,17 @@ public static async Task<BeatmapStore> CreateAsync(MySqlConnection connection, M
                 var workingBeatmap = new StreamedWorkingBeatmap(req.ResponseStream);
                 var calculator = ruleset.CreateDifficultyCalculator(workingBeatmap);
 
-                return calculator.Calculate(mods);
+                var attributes = calculator.Calculate(mods);
+
+                string[] tags =
```
peppy, line 98: Have to make sure to turn off indexing on these as soon as it goes live or we'll rack up a $1000 bill in no time

osu.Server.Queues.ScoreStatisticsProcessor/Stores/BeatmapStore.cs
```diff
@@ -107,12 +131,40 @@ public static async Task<BeatmapStore> CreateAsync(MySqlConnection connection, M
             return difficultyAttributes;
         }
 
+        /// <remarks>
+        /// This method attempts to create a simple solution to deciding if a <see cref="Mod"/> can be considered a ranked "legacy" mod.
+        /// Used by <see cref="GetDifficultyAttributesAsync"/> to decide if the current mod combination's difficulty attributes
+        /// can be fetched from the database.
+        /// </remarks>
+        private static bool isRankedLegacyMod(Mod mod) =>
```
peppy, line 139: I guess this is the real question of this PR is whether this is what we want. I'm pretty neutral on it, just have to add it to the list of things in our ever-increasing mental map of what needs to be updated when we make changes or add new mods.

I do agree with this listing legacy mods though. It's less likely that we'll add new mods to this list (will probably only be a consideration when we notice the pp calculations running slowly and need to force more performance out of things).

peppy (approved):

osu.Server.Queues.ScoreStatisticsProcessor/Stores/BeatmapStore.cs
bdach: I'm not even sure how to cross-check that this list is correct / complete right now.

Also I'd probably slightly prefer if this was listing leaf classes in the inheritance classes (don't like the `// this also catches ManiaModFadeIn` comment, and worried about future accidental inheritance), but probably not as important as _not being to tell if this is correct_.

osu.Server.Queues.ScoreStatisticsProcessor/Stores/BeatmapStore.cs
```diff
@@ -26,7 +32,7 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Stores
     /// </summary>
     public class BeatmapStore
     {
-        private static readonly bool use_realtime_difficulty_calculation = Environment.GetEnvironmentVariable("REALTIME_DIFFICULTY") != "0";
+        private static readonly bool prefer_realtime_difficulty_calculation = Environment.GetEnvironmentVariable("PREFER_REALTIME_DIFFICULTY") != "0";
```
bdach, line 35: I preferred the old naming of this envvar because this had `ALWAYS` in the name which feels more correct. This saying "prefer" gives me an impression that realtime will _sometimes_ be used - but no, that's not what this is - realtime will _always_ be used regardless of this envvar, it's just a question of _for which mods_ (with the flag off all mods that don't have db attributes well be realtimed, and with the flag on, _everything_ will).

tsunyoku: Yeah I think I agree with this. I think I changed it initially because I read the naming in the `BeatmapStore` (`use_realtime_difficulty_calculation`) and found `use` misleading because with these changes it might be used even if this is set to `0`. I can probably rename the env var back to `ALWAYS_REALTIME_DIFFICULTY` and make sure the name in `BeatmapStore` reflects that same naming.

bdach: `ALWAYS_USE_REALTIME_DIFFICULTY` if anything to appease the [grammar concern](https://github.com/ppy/osu-queue-score-statistics/pull/274#discussion_r1680417961).

I would have preferred that you wait for @peppy response to this review thread.

tsunyoku: Sure, sorry - I'll keep it on `ALWAYS_REALTIME_DIFFICULTY` for now to avoid dancing around more changes.

peppy: `ALWAYS_USE_REALTIME_DIFFICULTY` is fine by me. just force push the change if you want.

osu.Server.Queues.ScoreStatisticsProcessor/Stores/BeatmapStore.cs
```diff
@@ -107,12 +131,40 @@ public static async Task<BeatmapStore> CreateAsync(MySqlConnection connection, M
...
 
+        /// <remarks>
+        /// This method attempts to create a simple solution to deciding if a <see cref="Mod"/> can be considered a ranked "legacy" mod.
+        /// Used by <see cref="GetDifficultyAttributesAsync"/> to decide if the current mod combination's difficulty attributes
+        /// can be fetched from the database.
+        /// </remarks>
+        private static bool isRankedLegacyMod(Mod mod) =>
+            mod is ModNoFail
+                or ModEasy
+                or ModHidden // this also catches ManiaModFadeIn
```
bdach, line 142: > Also I'd probably slightly prefer if this was listing leaf classes in the inheritance classes (don't like the // this also catches ManiaModFadeIn comment, and worried about future accidental inheritance)

Case in point: this will _also_ catch `ManiaModCover`, [because `ManiaModCover : ManiaModWithPlayfieldCover : ModHidden`](https://github.com/ppy/osu/blob/7e5fda45396ce7823a7b5e63c7204c9ceafd184c/osu.Game.Rulesets.Mania/Mods/ManiaModCover.cs#L13).

osu.Server.Queues.ScoreStatisticsProcessor/Stores/BeatmapStore.cs
bdach: There are two places to check (that I won't link to, for hopefully obvious reasons). So at least I know that now.

bdach (changes requested):

osu.Server.Queues.ScoreStatisticsProcessor/Stores/BeatmapStore.cs
tsunyoku: Now doing more precise checks to aid against accidental inheritance.

osu.Server.Queues.ScoreStatisticsProcessor/Stores/BeatmapStore.cs
tsunyoku: Should be solved in https://github.com/ppy/osu-queue-score-statistics/pull/274/commits/09efc446d830650e0d28142b1c394058c64cfade.

peppy:
I think we can push this out for testing if @bdach is on board with the applied changes.

osu.Server.Queues.ScoreStatisticsProcessor/Stores/BeatmapStore.cs
bdach: You fixed the one case that was a problem but I still was not confident there won't be future problems with this so I pushed a test case (https://github.com/ppy/osu-queue-score-statistics/pull/274/commits/327669a1ec444100ade5444020d0fa9e44aac7e8) which should hopefully fail if someone inherits wrong in the future.

bdach:
Not sure how that happened but the envvar had a space in name. Removed in https://github.com/ppy/osu-queue-score-statistics/pull/274/commits/8b57913d86206f2ce686b8fb990550316201a9f9.

bdach:
Well I've done some testing and this appears to do what it says I guess. That said, @peppy If you're going to try this:

- **VERY IMPORTANT: Make sure to update the envvar from `REALTIME_DIFFICULTY` to `ALWAYS_USE_REALTIME_DIFFICULTY`**. It defaults to 1 (and did previously). On production it must currently be set to 0. If you don't update the envvar name (or probably, more likely, specify both in the interim), this is going to start using realtime difficulty for **ALL** scores which **WILL** die of death.
- The realtime difficulty process does a web request for **EVERY** score. There is **NO** caching instituted. This is not news, but since we're wanting to switch things on, this is a consideration that this PR introduces. Specifying `file://` protocol to bypass that **WILL NOT** work:

```
[network] 2024-08-07 09:33:59 [verbose]: Request to file:///home/dachb/Documents/opensource/diffcalc-sheet-generator/beatmaps/869310.osu failed with System.NotSupportedException: The 'file' scheme is not supported.
[network] 2024-08-07 09:33:59 [verbose]: at System.Net.Http.HttpClient.<SendAsync>g__Core|83_0(HttpRequestMessage request, HttpCompletionOption completionOption, CancellationTokenSource cts, Boolean disposeCts, CancellationTokenSource pendingRequestsCts, CancellationToken originalCancellationToken)
[network] 2024-08-07 09:33:59 [verbose]: at osu.Framework.IO.Network.WebRequest.internalPerform(CancellationToken cancellationToken).
```

- Realtime difficulty **WILL** immediately be used after this is deployed, for scores that use a few mods which did not exist in stable, and thus far worked because we basically went "this mod does not affect difficulty attributes, so we can pretend it does not exist for the sake of difficulty attribute lookup / treat it as nomod". As far as I know, this will be the case for mods such as **Muted** and **Traceable**. Possibly others that I'm forgetting.

bdach (approved):
Under the premises above

developomp:
If it's not a huge hassle, there should probably be a partial pp recalc for old lazer scores so they are retroactively rewarded pp.

bdach:
> If it's not a huge hassle, there should probably be a partial pp recalc for old lazer scores so they are retroactively rewarded pp.

We haven't done this thus far when switching mods to ranked, so I dunno about that.

stanriders:
> > If it's not a huge hassle, there should probably be a partial pp recalc for old lazer scores so they are retroactively rewarded pp.
> 
> We haven't done this thus far when switching mods to ranked, so I dunno about that.

It's probably gonna be covered by a pp deploy anyway, so it's not really necessary

peppy:
> Well I've done some testing and this appears to do what it says I guess. That said, @peppy If you're going to try this:
> 
> * **VERY IMPORTANT: Make sure to update the envvar from `REALTIME_DIFFICULTY` to `ALWAYS_USE_REALTIME_DIFFICULTY`**. It defaults to 1 (and did previously). On production it must currently be set to 0. If you don't update the envvar name (or probably, more likely, specify both in the interim), this is going to start using realtime difficulty for **ALL** scores which **WILL** die of death.
> * The realtime difficulty process does a web request for **EVERY** score (to fetch the `.osu` file). There is **NO** caching instituted. This is not news, but since we're wanting to switch things on, this is a consideration that this PR introduces. Specifying `file://` protocol to bypass that **WILL NOT** work:
> 
> ```
> [network] 2024-08-07 09:33:59 [verbose]: Request to file:///home/dachb/Documents/opensource/diffcalc-sheet-generator/beatmaps/869310.osu failed with System.NotSupportedException: The 'file' scheme is not supported.
> [network] 2024-08-07 09:33:59 [verbose]: at System.Net.Http.HttpClient.<SendAsync>g__Core|83_0(HttpRequestMessage request, HttpCompletionOption completionOption, CancellationTokenSource cts, Boolean disposeCts, CancellationTokenSource pendingRequestsCts, CancellationToken originalCancellationToken)
> [network] 2024-08-07 09:33:59 [verbose]: at osu.Framework.IO.Network.WebRequest.internalPerform(CancellationToken cancellationToken).
> ```
> 
>  * Realtime difficulty **WILL** immediately be used after this is deployed, for scores that use a few mods which did not exist in stable, and thus far worked because we basically went "this mod does not affect difficulty attributes, so we can pretend it does not exist for the sake of difficulty attribute lookup / treat it as nomod". As far as I know, this will be the case for mods such as **Muted** and **Traceable**. Possibly others that I'm forgetting.

Thanks for the summary. This all sounds in line with my understanding. I'll look to deploy this tomorrow.

smoogipoo:
Something that's been on my mind, but I've been weary of mentioning it because it'll get pushback ("release this ASAP and see if it breaks first!!!"), is that we already have https://github.com/ppy/osu-beatmap-difficulty-lookup-cache which is duplicating this functionality.

It may be worthwhile to hook that up (it's a web request that's being done anyway so overhead should be about the same) and optimising the caching local to that project.

tsunyoku:
> It may be worthwhile to hook that up (it's a web request that's being done anyway so overhead should be about the same) and optimising the caching local to that project.

I'm content with this, but it'll require adjustments over on that repository as it only caches for non-databased-attributes and is also configured to only compute attributes if `USE_DATABASE_LOOKUPS` isn't set. That repo however *does* have the advantage of it's currently caching when things are computed in real time, which might improve throughput for some common mod settings (mostly thinking of rate changes here) however I'm not sure if that would consume too much memory in the long-run as any settings combination would be new attributes to store. It'd also require moving over the logic to decide if a mod should use real time calculations (at least if we want it to function similarly to this PR) and I'm unsure if we want to tangle ranked-related logic (see `IsRankedLegacyMod`) into that project...

I think there's no harm in keeping things simple and deploying this PR as-is with the hopes of having a more centralised solution in the long-run. I imagine things will change with time anyways as lazer mods being ranked and awarding PP is a new concept in the first place.

tsunyoku:
Commenting as a means to ask if we can think about deploying this some time soon.

The main reasons I want this in are:

- It's a requirement to give rate changes PP
- It will allow us to make difficulty calculation changes that take into consideration the differences in mechanics between lazer and stable. For example: https://github.com/ppy/osu/pull/28267 & https://discord.com/channels/188630481301012481/542532479626772500/1367057400452087870
- Whilst I think we can/should rank Mirror without real time by treating it as no-mod (see https://github.com/ppy/osu/discussions/33076), real time would remove any need to have that debate
- Would allow other mods such as https://github.com/ppy/osu/pull/30232

I'm not suggesting the immediate ranking of mods following the deploy of this. However, I'd like to see them ranked in a reasonable timeline and we can't really do that without knowing that this works well in production.

peppy:
> It will allow us to make difficulty calculation changes that take into consideration the differences in mechanics between lazer and stable

This would imply you are intending to use this for either all lazer or all stable scores, which is a much larger scope than the odd mod where it's not in the database. Can you confirm this is the case? If so, there are still some prerequisites to make this work.

tsunyoku:
Hmm, you are right. It would be for all lazer scores, since we have the database tables for legacy scores.

Somehow that much slipped my mind, and this PR does not cause all lazer scores to go down the realtime path because it was intentionally built that way.

I agree that this would require further work to run on absolutely every score, so I'm fine if that is an improvement which comes later - being able to rank lazer mods which have intentional star rating differences is already a big step forward and this PR in its current state allows for that.

NokoShika:
I understand that this is difficult to implement without knowing how many bytes it will take to index all of this, but LPU will at least save disk space when it starts to fill up. I don’t have access to the backend and I don’t know what’s going on there like the back of my hand, and I’m not very knowledgeable about this, so consider this purely my idea.

tsunyoku:
Was reminded of this PR so fixed conflicts. Realised in doing so that the method I was using to deduce whether to use real-time calculations was flawed.

The old method was checking if the mod was either a non-legacy mod, or a legacy mod with non-default settings. This sounds fine at first glance. However we have several ranked lazer mods with settings that don't affect difficulty, and several legacy mods with non-default settings that don't affect difficulty. Some examples:

- Muted
- Accuracy Challenge
- Double Time (with Adjust Pitch enabled)

With the old logic, these would be sent down realtime calculations for _no_ reason.

Simplest way I could think of to solve this is just to explicitly whitelist which mods warrant realtime calculations.

I've created a simple method that enables realtime for a limited number of scenarios that I'd initially be interested in seeing this change for - this is for demonstrative purposes, the actual choices made aren't that important. Every mod in this new method is currently unranked, meaning it will never reach this method as the `ScorePerformanceProcessor` will exit early. Hopefully this makes deploying this change easy, as it currently won't send _any_ scores down the realtime calculation path.

Let me know if you see issues with this, but I think this is the best way to do it.

PR #280, opened 2024-10-18
Use lightweight models to avoid `Guid` overheads

peppy:


peppy:
Well this definitely broke something, thanks tests ❤️

bdach:
> Well this definitely broke something

No, not sure it did. What happened is that all the diff/ppcalc changes broke stuff or changed values and thus the tests went red.

Can be verified by local checkout of game at ppy/osu@2024.911.0 (which looks to be a... deleted tag, but points to https://github.com/ppy/osu/commit/02f8f226f845a6a0fe51f7f03cd3d7a9945f1943 on my copy of the repository), and then cherry-picking https://github.com/ppy/osu/commit/8193038986d2dc205d83568ff2146220bc203dc5 on top of that. Tests will be green again.

Pushed https://github.com/ppy/osu-queue-score-statistics/pull/280/commits/715606461fbdb468b235a32da7418bed9431e7eb to fix the failures.

bdach (approved):

peppy:
Hmm, I thought that was the case at first but one of the failures looked grim:

Safari 2024-10-21 at 09 41 51

I guess this was a red herring, phew.

PR #282, opened 2024-10-18
Cache finalised attributes to avoid conversion overhead every usage

peppy:


osu.Server.Queues.ScoreStatisticsProcessor/Stores/BeatmapStore.cs
```diff
@@ -82,29 +82,34 @@ public static async Task<BeatmapStore> CreateAsync(MySqlConnection connection, M
...
 
-            return difficultyAttributes;
+                difficultyAttributes = LegacyRulesetHelper.CreateDifficultyAttributes(ruleset.RulesetInfo.OnlineID);
+                difficultyAttributes.FromDatabaseAttributes(dbAttribs.ToDictionary(a => (int)a.attrib_id, a => (double)a.value), beatmap);
+                return difficultyAttributes;
+            }
+            finally
+            {
+                attributeCache[key] = difficultyAttributes;
+            }
```
bdach, line 112: I believe all of this works but it's intricate enough to be pushing my limits of how much subtlety I'm willing to accept in code, what with all of the juggling of `difficultyAttributes`, implicit null returns of `TryGetValue()`, and `finally` used to coalesce return paths in order to populate `attributeCache`...

bdach (approved):

osu.Server.Queues.ScoreStatisticsProcessor/Stores/BeatmapStore.cs
peppy: Hmm, this is the standard way I'd write any caching code (I believe it's done this was elsewhere) but I'm neutral to approaching it differently. Would be be more understandable if the early return did the null-caching inline?

bdach: The general pattern is standard yes, I'm more talking about the mechanics here.

> Would be be more understandable if the early return did the null-caching inline?

To a degree, but it would also make it mechanically uglier, so I decided I would just put up with this as is.

smoogipoo: One thing to keep in mind is that this will cache attributes even if they fail, and then they will pass as if nothing went wrong the next time around. There's two places that can fail - either the attributes are not present in the DB which I can't immediately think of a situation in which it should be allowed, or the deserialisation fails which is always a bad sign.

PR #283, opened 2024-10-18
Use static ruleset classes rather than constructing on each usage

peppy:
These are used a *lot*, so this is quite a large saving. Mainly just aiming to get GC time down.


bdach (approved):
Seems safe

PR #284, opened 2024-10-21
Score processing command updates for upcoming (and future PP deploys)

peppy:
This adds a new command which processes all scores in the database consecutively (based on primary key). This is more efficient for both reads and write due to the table's compression setting. In practice it's over a magnitude faster, and will replicate better.

In additions, this PR also:

- Adds better console output of the progress of ongoing operations
- Removes a stray `AsParallel()` call that didn't need to exist
- Allows cancellation during a single user processing
- Allows bypassing beatmap store refreshes for performance reasons
- Avoids database writes when the existing pp value is already correct-enough (within `0.1` as per [what the spreadsheet process currently checks for](https://discord.com/channels/90072389919997952/1083320217259749376/1296763588433674260))

Without database writes, this now processes around 120-180k scores per second with 8 threads. Increasing thread count further does not seem to help much, but this is already way faster than we can write to the database so it's fine.

With database writes,  increasing the thread count to 32 performs better (more throughput at the database end). We are still capped around 16k/s, which is about the fastest we've been able to `INSERT` to the new table. It should be fast enough for our purposes.

Only unknown is how replication will respond to the load of running a full reprocess.

Example of console output, 32 threads with writes enabled:

https://github.com/user-attachments/assets/0d86a30c-7000-49c9-b0e1-f9adddc0c3e0

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Performance/PerformanceCommand.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Performance/PerformanceCommand.cs
@@ -103,7 +103,6 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Performance
             await Task.WhenAll(Partitioner
                                .Create(values)
                                .GetPartitions(Threads)
-                               .AsParallel()
                                .Select(processPartition));
 
             async Task processPartition(IEnumerator<T> partition)
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Performance/PerformanceCommand.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Performance/PerformanceCommand.cs
@@ -92,7 +92,7 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Performance
             await ProcessPartitioned(userIds, async userId =>
             {
                 using (var db = DatabaseAccess.GetConnection())
-                    await ScoreProcessor.ProcessUserScoresAsync(userId, RulesetId, db);
+                    await ScoreProcessor.ProcessUserScoresAsync(userId, RulesetId, db, cancellationToken: cancellationToken);
 
                 Console.WriteLine($"Processed {Interlocked.Increment(ref processedCount)} of {userIds.Length}");
             }, cancellationToken);
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Performance/Scores/UpdateAllScoresByUserCommand.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Performance/Scores/UpdateAllScoresByUserCommand.cs
@@ -2,6 +2,7 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System;
+using System.Data;
 using System.Diagnostics;
 using System.Linq;
 using System.Threading;
@@ -81,7 +82,7 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Performance.Scores
                 await ProcessPartitioned(userIds, async userId =>
                 {
                     using (var db = DatabaseAccess.GetConnection())
-                    using (var transaction = await db.BeginTransactionAsync(cancellationToken))
+                    using (var transaction = await db.BeginTransactionAsync(IsolationLevel.ReadUncommitted, cancellationToken))
                     {
                         Interlocked.Add(ref totalScores, (ulong)await ScoreProcessor.ProcessUserScoresAsync(userId, RulesetId, db, transaction, cancellationToken));
                         await transaction.CommitAsync(cancellationToken);
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Performance/Scores/UpdateAllScoresByUserCommand.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Performance/Scores/UpdateAllScoresByUserCommand.cs
@@ -2,6 +2,7 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System;
+using System.Diagnostics;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
@@ -15,7 +16,7 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Performance.Scores
     [Command(Name = "all-users", Description = "Computes pp of all scores from all users.")]
     public class UpdateAllScoresByUserCommand : PerformanceCommand
     {
-        private const int max_users_per_query = 10000;
+        private const int max_users_per_query = 1000;
 
         [Option(Description = "Continue where a previously aborted 'all' run left off.")]
         public bool Continue { get; set; }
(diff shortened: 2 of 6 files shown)


osu.Server.Queues.ScoreStatisticsProcessor/Commands/Performance/PerformanceCommand.cs
```diff
@@ -103,7 +103,6 @@ protected async Task ProcessPartitioned<T>(IEnumerable<T> values, Func<T, Task>
             await Task.WhenAll(Partitioner
                                .Create(values)
                                .GetPartitions(Threads)
-                               .AsParallel()
```
peppy, line 106: I am not 100% clear about *when* you're supposed to use `AsParallel` but it only seems to be required to make synchronous enumerables into asynchronous ones, which is not what we need here.

It has no effect on processing performance at all.

smoogipoo: I'm very confused. Why is this not what we want?

Here's a simple console app:
```csharp
using System.Collections.Concurrent;
using System.Diagnostics;

Stopwatch sw = Stopwatch.StartNew();

await processPartitioned(Enumerable.Range(0, 1000), t =>
{
    Thread.Sleep(100);
    return Task.CompletedTask;
}, CancellationToken.None);

sw.Stop();
Console.WriteLine(sw.ElapsedMilliseconds);

async Task processPartitioned<T>(IEnumerable<T> values, Func<T, Task> processFunc, CancellationToken cancellationToken)
{
    await Task.WhenAll(Partitioner
                       .Create(values)
                       .GetPartitions(8)
                       .AsParallel()
                       .Select(processPartition));

    async Task processPartition(IEnumerator<T> partition)
    {
        using (partition)
        {
            while (partition.MoveNext())
            {
                if (cancellationToken.IsCancellationRequested)
                    return;

                await processFunc(partition.Current);
            }
        }
    }
}
```
With the `AsParallel()`, this takes around 13s to complete, which is expected given the delay. Without, it does not complete for a long time.

peppy: You're using a synchronous (non-`async`) delegate function there, which is where the difference lies. It's not required for `async` functions.

ie. try replacing the `Thread.Sleep` with an `await Task.Delay`.

smoogipoo: Right, sure. But isn't that what we want?

That is, regardless of if my delegate is:
```
Console.WriteLine(t);
Thread.Sleep(100);
```
or
```
Console.WriteLine(t);
await Task.Delay(100);
```

I'm getting the same throughput with `AsParallel()`, whereas not without it. Did you run into some sort of issue that made you change this or something?

peppy: We are using `async` everywhere, so it's just not necessary. That's all.

smoogipoo: 🙈

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Performance/Scores/UpdateAllScoresCommand.cs
```diff
@@ -0,0 +1,120 @@
...
+                Console.WriteLine($"id: {currentScoreId:N0} changed: {changedPp:N0} ({processedCount:N0} of {lastScoreId:N0} {(float)processedCount / lastScoreId:P1}) {rate:N0}/s");
+            }
+
+            return 0;
+
+            Task<List<SoloScore>> getNextScores() => Task.Run(() =>
+            {
+                List<SoloScore> scores = new List<SoloScore>();
+
+                scores.Clear();
```
smoogipoo, line 111: Seems unnecessary?

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Performance/Scores/UpdateAllScoresCommand.cs
```diff
@@ -0,0 +1,120 @@
...
+
+                sw.Restart();
+
+                var scores = await nextScores;
+                nextScores = getNextScores();
+
+                if (scores.Count == 0)
+                    break;
+
+                await Task.WhenAll(Partitioner.Create(scores).GetPartitions(Threads).Select(async partition =>
```
smoogipoo, line 71: Any reason for not using `ProcessPartitioned`?

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Performance/Scores/UpdateAllScoresCommand.cs
```diff
@@ -0,0 +1,120 @@
...
+                {
+                    using (var connection = DatabaseAccess.GetConnection())
+                    using (var transaction = await connection.BeginTransactionAsync(IsolationLevel.ReadUncommitted, cancellationToken))
+                    using (partition)
+                    {
+                        while (partition.MoveNext())
+                        {
+                            if (cancellationToken.IsCancellationRequested)
+                                return;
```
smoogipoo, line 80: Note that if this is cancelled, then you're going to be printing a `currentScoreId` that may not have actually been processed yet. `ProcessPartitioned` has the same issue.

You may want to check the cancellation after the `await Task.WhenAll()` and not continue printing, lest there be confusion and scores missed because it was restarted from the wrong point.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Performance/Scores/UpdateAllScoresCommand.cs
```diff
@@ -0,0 +1,120 @@
...
+
+                currentScoreId = scores.Last().id;
+
+                if (rate == 0)
+                    rate = ((double)scores.Count / sw.ElapsedMilliseconds * 1000);
+                else
+                    rate = rate * 0.95 + 0.05 * ((double)scores.Count / sw.ElapsedMilliseconds * 1000);
+
+                Console.WriteLine(ScoreProcessor.BeatmapStore?.GetCacheStats());
+                Console.WriteLine($"id: {currentScoreId:N0} changed: {changedPp:N0} ({processedCount:N0} of {lastScoreId:N0} {(float)processedCount / lastScoreId:P1}) {rate:N0}/s");
```
smoogipoo, line 102: Nit: `{count} of {lastScoreId}` is not really correct but I guess it's hard to do much better. Maybe `of {lastScoreId - From}` could be tried but /shrug.

smoogipoo (changes requested):

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Performance/Scores/UpdateAllScoresCommand.cs
peppy: I couldn't make the transaction work easily when using that, so had to inline it.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Performance/Scores/UpdateAllScoresCommand.cs
peppy: Yeah, it's kinda best-effort. I'll make it read better.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Performance/Scores/UpdateAllScoresCommand.cs
peppy: https://github.com/ppy/osu-queue-score-statistics/pull/284/commits/1cb764f43ef74eb1e591014900cb31e9210052e4

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Performance/Scores/UpdateAllScoresCommand.cs
peppy: https://github.com/ppy/osu-queue-score-statistics/pull/284/commits/92700c1d3ada0c3ee69951d80d74d869ffccb254

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Performance/Scores/UpdateAllScoresCommand.cs
peppy: 4fdf3a9e9bb7fa8fe36bae993385a6d33d097163

JetBrains Rider-EAP 2024-10-21 at 10 55 49

No real point showing the max ID so I just removed that and modified the % part as you proposed.

smoogipoo (approved):

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Performance/Scores/UpdateAllScoresByUserCommand.cs
```diff
@@ -12,76 +14,92 @@
 
 namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Performance.Scores
 {
-    [Command(Name = "all", Description = "Computes pp of all scores from all users.")]
+    [Command(Name = "all-users", Description = "Computes pp of all scores from all users.")]
```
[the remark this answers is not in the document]
smoogipoo: Am running a test on it in https://github.com/ppy/osu/pull/27691, but won't wait for completion before merging.

smoogipoo (approved):
lgtm

PR #289, opened 2024-10-29
Unapply non-buffered performance optimisation

peppy:
- [x] Depends on https://github.com/ppy/osu-queue-score-statistics/pull/288

ProxySQL doesn't work great with this. If there's a pause (or sometimes even if there isn't) ie. connection may be dropped causing the reading to fail and abort the whole process.

I'm PRing this separately because I may bring this back in the near future, depending on how things go. Right now we're bottlenecked on writes, not reads, so it's not a huge issue.

Already using on production run.


peppy:
Don't merge this quite yet. It's quite a performance hit for certain user cases and I might find a better way.

peppy:
@smoogipoo this is good to go for now

smoogipoo (approved):

PR #292, opened 2024-10-31
Improve throughput of other commands using `ProcessPartitioned` by allowing transaction and connection sharing

peppy:
Some more optimisations I made in order to make updating user totals as a batch more efficient.

More details in some commit messages.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Performance/Scores/UpdateAllScoresByUserCommand.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Performance/Scores/UpdateAllScoresByUserCommand.cs
@@ -2,7 +2,6 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System;
-using System.Data;
 using System.Diagnostics;
 using System.Linq;
 using System.Threading;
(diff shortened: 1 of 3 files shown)


osu.Server.Queues.ScoreStatisticsProcessor/Commands/Performance/Scores/UpdateAllScoresByUserCommand.cs
```diff
@@ -78,14 +77,10 @@ protected override async Task<int> ExecuteAsync(CancellationToken cancellationTo
...
+                await ProcessPartitioned(userIds, async (conn, transaction, userId) =>
                 {
-                    using (var partitionDb = DatabaseAccess.GetConnection())
-                    using (var transaction = await partitionDb.BeginTransactionAsync(IsolationLevel.ReadUncommitted, cancellationToken))
-                    {
-                        Interlocked.Add(ref totalScores, (ulong)await ScoreProcessor.ProcessUserScoresAsync(userId, RulesetId, db, transaction, cancellationToken));
-                        await transaction.CommitAsync(cancellationToken);
-                    }
+                    Interlocked.Add(ref totalScores, (ulong)await ScoreProcessor.ProcessUserScoresAsync(userId, RulesetId, conn, transaction, cancellationToken));
+                    await transaction.CommitAsync(cancellationToken);
```
smoogipoo, line 83: Is this intentional? It sounds like this would break things because the transaction would still be in use for the next iteration.

peppy: looks like a merge resolution fail

smoogipoo (approved):

PR #297, opened 2024-11-13
Fix various content type issues on LIO request helpers

bdach:
As [pointed out internally](https://discord.com/channels/90072389919997952/983550677794050108/1306327452364177418).

- `plain/text` is exactly backwards; [it's actually `text/plain`](https://developer.mozilla.org/en-US/docs/Web/HTTP/MIME_types#textplain).
- The MIME type spec is redundant there anyways. [`StringContent` implicitly means `text/plain` by default until specified otherwise.](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.stringcontent.-ctor?view=net-8.0#system-net-http-stringcontent-ctor(system-string))
- And the only place that used the `StringContent` path was wrong anyway, it was supposed to be doing `application/x-www-form-urlencoded`. [Which `FormUrlEncodedContent` is for, and is now used in the relevant context.](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.formurlencodedcontent?view=net-8.0)

Mind you, all this seems to have absolutely no impact on anything whatsoever and be 100% academical, because as far as I can see, on local environment the web LIO endpoint just responds with 204 anyways, with or without this PR. I'd expect 415 if it didn't like anything above, so I'm wagering it just didn't care or check and all of this had an impact of exactly none whatsoever.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Helpers/LegacyDatabaseHelper.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Helpers/LegacyDatabaseHelper.cs
@@ -2,6 +2,7 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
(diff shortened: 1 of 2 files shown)


nanaya:
the parameters are ignored when they're invalid so it has been a noop.

also might as well just pass it as json and remove support for the other type...

bdach:
> also might as well just pass it as json and remove support for the other type...

Sorry, I'm going to ask for clarification, as this is ambiguous to me. Are you saying that the lio endpoint in question supports json? And thus we can just send json in a similar shape?

nanaya:
apart of the oauth, I think they all support all parameter types (as long it's valid)

nanaya:
oh and while at it maybe don't customize the request content if the post object is null...

peppy:
@bdach here's an example of json usage: https://github.com/ppy/osu-ass/blob/289397bd71f296eed1cfc92d54e9f6c01a532d20/osu.Server.AutobahnSurveillanceSystem/User.cs#L834

peppy (approved):

bdach:
I was gonna still address the points above, but I guess now that this is merged I... won't? Should have marked as blocked I guess :shrug:

bdach:
Alright, I'll do a round 2 then.

PR #299, opened 2024-11-18
Adjust LIO request helper further

bdach:
Follow up from https://github.com/ppy/osu-queue-score-statistics/pull/297

- Removed support for both `text/plain` (completely unnecessary ever, web can't do that) and `application/x-www-form-urlencoded` (web _can_ do that but why leave 2 formats in when 1 can work), leaving only `application/json`
- If the request is made with a null payload object, do not set `Content-Type`


bdach:
Test failure should be unrelated...?

peppy (approved):

PR #300, opened 2024-11-18
Split domain to which LIO requests are directed to envvar

bdach:
On several occasions, I've tried to test something involving LIO full-stack and forgot that the LIO helper hardcodes production. Thankfully I never did damage due to the interop secret check, but it's a bit bone-chilling every time. Additionally I believe this situation can't be correct as far as staging is involved, either.

The envvar is `LEGACY_IO_DOMAIN` (open to alternative naming suggestions). Same as the shared interop secret, this envvar is optional in debug (and it missing causes LIO to become a no-op), and required with no default and hard-fail on missing in release.

cc @ThePooN as requested (for presumed deployment config updates)

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,9 +103,19 @@ Defaults to `localhost`.
 
 Enables DataDog origin detection when running in a container. See [DataDog documentation](https://docs.datadoghq.com/developers/dogstatsd/?tab=kubernetes&code-lang=dotnet#origin-detection-over-udp).
 
+### LEGACY_IO_DOMAIN
+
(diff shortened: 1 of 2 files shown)


peppy (approved):

ThePooN:
mmh, perhaps should've re_used `APP_URL` from osu-web?

bdach:
I guess, although to me `APP_URL` seems sorta... non-specific? Can change if you think it's worth the effort...

ThePooN:
nah, dont think its worth it. on a second thought, it would be a confusing one to override for internal routing as well.

PR #302, opened 2024-11-20
Fix multiple issues related to mania key count stats population

### LINKED ISSUE #295: User total processing commands don't update mania key-mode tables
Processing user totals runs the `UserTotalPerformanceProcessor`:

https://github.com/ppy/osu-queue-score-statistics/blob/3ac03768bb93cb2568d1b87f02c5c35953b48c74/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Performance/PerformanceCommand.cs#L75

But not the [`ManiaKeyModeUserStatsProcessor.cs`](https://github.com/ppy/osu-queue-score-statistics/blob/3ac03768bb93cb2568d1b87f02c5c35953b48c74/osu.Server.Queues.ScoreStatisticsProcessor/Processors/ManiaKeyModeUserStatsProcessor.cs).

peppy:
Opening this for self testing / review

- Closes #295
- [x] Make sure this works as expected.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Performance/PerformanceCommand.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Performance/PerformanceCommand.cs
@@ -86,12 +86,20 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Performance
 
                         if (keyModeStats != null)
                         {
+                            double rankScoreBefore = keyModeStats.rank_score;
+                            double accBefore = keyModeStats.accuracy_new;
+
                             await ManiaKeyModeProcessor.UpdateUserStatsAsync(keyModeStats, keyCount, db, transaction, updateIndex: false);
-                            await db.ExecuteAsync(
-                                $"UPDATE `{keyCountTableName}` "
-                                + $"SET `rank_score` = @rank_score, `playcount` = @playcount, `rank_score_index` = @rank_score_index, `accuracy_new` = @accuracy_new, "
-                                + $"`x_rank_count` = @x_rank_count, `xh_rank_count` = @xh_rank_count, `s_rank_count` = @s_rank_count, `sh_rank_count` = @sh_rank_count, `a_rank_count` = @a_rank_count "
-                                + $"WHERE `user_id` = @user_id", keyModeStats, transaction);
+
+                            if (Math.Abs(rankScoreBefore - keyModeStats.rank_score) > 0.1 ||
+                                Math.Abs(accBefore - keyModeStats.accuracy_new) > 0.1)
+                            {
+                                await db.ExecuteAsync(
+                                    $"UPDATE `{keyCountTableName}` "
+                                    + $"SET `rank_score` = @rank_score, `playcount` = @playcount, `rank_score_index` = @rank_score_index, `accuracy_new` = @accuracy_new, "
+                                    + $"`x_rank_count` = @x_rank_count, `xh_rank_count` = @xh_rank_count, `s_rank_count` = @s_rank_count, `sh_rank_count` = @sh_rank_count, `a_rank_count` = @a_rank_count "
+                                    + $"WHERE `user_id` = @user_id", keyModeStats, transaction);
+                            }
                         }
                     }
                 }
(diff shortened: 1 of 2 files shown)


peppy:
This should be good for review now. I'm running this in production as a final test run (after manually testing on users in both active and inactive state, confirming that active users see no-change and inactive users see their pp/acc getting updated).

I did take this opportunity to make the logic conform close to the main user stats processor, so the diff might be a bit hard to read. Just viewing the few files directly and ignoring the diff, treating as if it's a new submission, may work better.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/ManiaKeyModeUserStatsProcessor.cs
```diff
@@ -94,24 +62,75 @@ public void ApplyToUserStats(SoloScore score, UserStats fullRulesetStats, MySqlC
...
+                conn.Execute(
+                    $"UPDATE `{keyCountTableName}` "
+                    + $"SET `rank_score` = @rank_score, `playcount` = @playcount, `rank_score_index` = @rank_score_index, `accuracy_new` = @accuracy_new, "
+                    + $"`x_rank_count` = @x_rank_count, `xh_rank_count` = @xh_rank_count, `s_rank_count` = @s_rank_count, `sh_rank_count` = @sh_rank_count, `a_rank_count` = @a_rank_count "
+                    + $"WHERE `user_id` = @user_id", keyModeStats, transaction);
             }
             else
             {
-                conn.Execute($"UPDATE `{keyCountTableName}` SET `playcount` = `playcount` + 1 WHERE `user_id` = @user_id", keymodeStats, transaction);
+                conn.Execute($"UPDATE `{keyCountTableName}` SET `playcount` = @playcount, `last_played` = NOW() WHERE `user_id` = @user_id", keyModeStats, transaction);
```
bdach, line 93: Why is `last_played` seemingly only updated in this branch? Shouldn't it also be getting set in the `score.preserve` branch?

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Performance/PerformanceCommand.cs
```diff
@@ -69,6 +73,37 @@ await ProcessPartitioned(userIds, async (db, transaction, userId) =>
...
+                            double accBefore = keyModeStats.accuracy_new;
+
+                            await ManiaKeyModeProcessor.UpdateUserStatsAsync(keyModeStats, keyCount, db, transaction, updateIndex: false);
+
+                            if (Math.Abs(rankScoreBefore - keyModeStats.rank_score) > 0.1 ||
+                                Math.Abs(accBefore - keyModeStats.accuracy_new) > 0.1)
+                            {
+                                await db.ExecuteAsync(
+                                    $"UPDATE `{keyCountTableName}` "
+                                    + $"SET `rank_score` = @rank_score, `playcount` = @playcount, `rank_score_index` = @rank_score_index, `accuracy_new` = @accuracy_new, "
```
bdach, line 99: Why is `rank_score_index` in this update if the `UpdateUserStatsAsync()` call prior was with `updateIndex: false`, which means that in theory the rank wouldn't have been touched...?

And, as a corollary to the above, what will ensure that `rank_score_index` will be eventually correct after this is run? The daily rank calculation cron I guess?

peppy: > And, as a corollary to the above, what will ensure that rank_score_index will be eventually correct after this is run? The daily rank calculation cron I guess?

That's the one, yep. I'm doing it this way in line with the (general) user totals update command.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/ManiaKeyModeUserStatsProcessor.cs
peppy: good catch, got lost in one of my refactors.

peppy: Fixed via 83bea01

bdach (approved):
seems ok

PR #303, opened 2024-11-21
Enable user rank processor on legacy scores

peppy:
A step towards sanity, with a hope of resolving https://github.com/ppy/osu-web/issues/8443. Will be deployed alongside https://github.com/peppy/osu-web-10/commit/0233c980bd2f201448781068d842908c04a28e18.


bdach:
As remarked elsewhere, the same class of bug applies to ranked score, so we'd probably want to apply the same process to `RankedScoreProcessor` too at some point. Not sure if that point is now or not.

Otherwise, I don't have any objections.

peppy:
I'll be turning these on one at a time just, until we have them all on.

bdach (approved):

PR #304, opened 2024-11-21
Add command to fix non-matching rank counts

peppy:
Inspiration taken from [web-10 implementation](https://github.com/peppy/osu-web-10/blob/master/www/web/adm-fixrankcounts.php), but updated for the new scores table.

I haven't yet tested this (still have an ongoing processing queue which I don't want to run in parallel). Just PRing as I think it's mostly ready to go. I'll test it tomorrow and make any required amendments.

In conjunction with https://github.com/ppy/osu-queue-score-statistics/pull/303 (already deployed), closes https://github.com/ppy/osu-web/issues/8443.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/VerifyUserRankCounts.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/VerifyUserRankCounts.cs
@@ -1,0 +1,155 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Dapper;
+using McMaster.Extensions.CommandLineUtils;
+using MySqlConnector;
+using osu.Game.Scoring;
+using osu.Server.QueueProcessor;
+using osu.Server.Queues.ScoreStatisticsProcessor.Helpers;
+using osu.Server.Queues.ScoreStatisticsProcessor.Models;
+
+namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Maintenance
+{
+    [Command("verify-user-rank-counts", Description = "Verifies SS/S/A rank counts for all users")]
+    public class VerifyUserRankCounts
+    {
+        [Option("--sql", Description = "Specify a custom query to limit the scope of pumping")]
+        public string? CustomQuery { get; set; }
+
+        /// <summary>
+        /// The ruleset to run this verify job for.
+        /// </summary>
+        [Option(CommandOptionType.SingleValue, Template = "--ruleset-id")]
+        public int RulesetId { get; set; }
+
+        [Option(CommandOptionType.SingleOrNoValue, Template = "-v|--verbose", Description = "Output verbose information on processing.")]
+        public bool Verbose { get; set; }
+
+        [Option(CommandOptionType.SingleOrNoValue, Template = "--dry-run")]
+        public bool DryRun { get; set; }
+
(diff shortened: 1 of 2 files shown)


osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/VerifyUserRankCounts.cs
```diff
@@ -0,0 +1,161 @@
...
+                    { ScoreRank.S, 0 },
+                    { ScoreRank.SH, 0 },
+                    { ScoreRank.X, 0 },
+                    { ScoreRank.XH, 0 },
+                };
+
+                if (Verbose)
+                    Console.WriteLine($"Processing user {userId} ({scores.Count()} scores)..");
+
+                IEnumerable<SoloScore?> maxScoresByBeatmap = scores.GroupBy(s => s.beatmap_id).Select(g => g.MaxBy(s => s.total_score));
```
bdach, line 97: The processor also orders the scores by ID descending in case of a total score tie:

https://github.com/ppy/osu-queue-score-statistics/blob/fc43f221bbc46e547346ec7827dbb92b1995263c/osu.Server.Queues.ScoreStatisticsProcessor/Helpers/DatabaseHelper.cs#L182

Not sure this is gonna matter in practice, but there's a chance it might, probably higher than I think. Not sure what this `.MaxBy()` is gonna do in such a case. Wouldn't be surprised if it didn't do the right thing (and by "right" I mean "the same thing that the rank count processor does").

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/VerifyUserRankCounts.cs
```diff
@@ -0,0 +1,161 @@
...
+                        Console.WriteLine($"x:  {userStats.x_rank_count} ->  {counts[ScoreRank.X]}");
+                        Console.WriteLine($"xh: {userStats.xh_rank_count} ->  {counts[ScoreRank.XH]}");
+                        Console.WriteLine();
+                    }
+
+                    if (!DryRun)
+                        await DatabaseHelper.UpdateUserStatsAsync(userStats, db, transaction);
+                }
+
+                await transaction.CommitAsync(cancellationToken);
```
bdach, line 157: Bit unfortunate that even in dry run mode this is gonna hold a transaction open for the entire duration of the operation even despite issuing no sql commands. Also a commit in a dry run mode is a tad scary.

In principle all of this is safe and good as written, but I do wonder if things can't be reshuffled so that dry run never takes out a txn in the first place.

peppy: I can just make it rollback if in dryrun if that is enough to make you feel more at ease?

bdach: If it works for you it works for me. That part wasn't the one I was worried about so much, the part that worries me more is unnecessary transaction contention when running this in dry run (although the devil's advocate take on that is that *incurring* this transaction contention will be useful to check if things don't fall over too hard on the *real* run).

peppy: Yeah, I'd want the transaction to remain even in a dry run for that purpose too.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/VerifyUserRankCounts.cs
peppy: That's a really weird choice (to use the most recent score) but I guess it's kinda required in this case for some kind of sanity so I'll just go with it.

bdach (approved):
passes the eye test, going to leave this hanging approved pending testing

peppy:
Seems to be working, going to merge for a more efficient production run.

PR #308, opened 2024-12-20
Add new medals

bdach:


peppy (approved):
On a basic check this all looks good

Issue #314, opened 2025-01-10
Remove preserve flag set


MaxTYT-skurr:
While looking through the code, I found the following snippet:
https://github.com/ppy/osu-queue-score-statistics/blob/2523de21d0c7e9661f664884e7d0bc9ab3eda268/osu.Server.Queues.ScoreStatisticsProcessor/ScoreStatisticsQueueProcessor.cs#L239-L246
The linked issue seems to be resolved with 2 PRs having been merged, meaning the snippet would now be unnecessary according to the comment. Just thought I'd give a heads up. 
Kind regards

PR #315, opened 2025-01-14
Stop marking preserve on new score arriving

peppy:
Closes https://github.com/ppy/osu-queue-score-statistics/issues/314.

Leaving logging so we have visibility if anything goes awry.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/ScoreStatisticsQueueProcessor.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/ScoreStatisticsQueueProcessor.cs
@@ -236,17 +236,11 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor
                         transaction.Commit();
                     }
 
-                    // TODO: this can be removed after https://github.com/ppy/osu-web/issues/10942 is closed out.
-                    // Intentionally not part of the transaction to avoid deadlocks.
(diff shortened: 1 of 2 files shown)


smoogipoo (approved):

osu.Server.Queues.ScoreStatisticsProcessor/ScoreStatisticsQueueProcessor.cs
```diff
@@ -236,17 +236,11 @@ protected override void ProcessResult(ScoreItem item)
...
-                    {
-                        // For now, just assume all passing scores are to be preserved.
-                        conn.Execute("UPDATE scores SET preserve = 1 WHERE id = @Id", new { Id = score.id });
-                    }
-
                     foreach (var p in enumerateValidProcessors(score))
                         p.ApplyGlobal(score, conn);
+
+                    if (!score.preserve)
+                        Console.WriteLine($"Score {score.id} was processed but not preserved");
```
bdach, line 243: Only passed scores are initially marked as preserved, and we do process failed scores, although only partially. So this is going to start spamming immediately after deploy for every failed score, is it not? If the goal of this was to be a bulwark against expectations not being met, then it kinda doesn't work...

I'd expect something closer to

```suggestion
                    if (score.passed && !score.preserve)
                        Console.WriteLine($"Passed score {score.id} was processed but not preserved");
```

bdach (approved):

PR #320, opened 2025-01-23
Add command to process all scores

peppy:
Similar to `PumpAllScores` but without the queue.

Useful for local testing.


smoogipoo (approved):
As mentioned IRL I'm a little bit sketched out by rerunning processors. I think I brought up a reason why we should never do it in general at some point in the past - lossy data perhaps? So I would advise caution.

peppy:
It's okay it's totally just for local testing

PR #323, opened 2025-02-27
Update packages

peppy:
I've made tests pass, but not 100% if we need to add the now missing attributes back to beatmap (or if it doesn't really matter because the tests aren't focused on that in the first place).

Specific attention to the taiko test, I guess (cc @bdach).

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,0 +1,7 @@
+{
+  "sdk": {
+    "version": "8.0.100",
+    "rollForward": "latestFeature",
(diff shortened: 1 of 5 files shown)


bdach (approved):
As far as I can tell the taiko detection thing exercised by the test mentioned is still working correctly despite the values changing (can be exercised by adjusting the beatmap in that test to another ruleset and observing the subsequent test failure).

PR #325, opened 2025-03-06
Stop catching exceptions in `ScorePerformanceProcessor`

peppy:
It seems like this is occasionally failing, but due to exceptions being caught and logged to console, we have not seen an actual failure.


bdach:
It does appear that command usages will change semantics with this change, from skipping over failed items to hard-dying on them. Which might not be a bad thing? Dunno.

There's even a seemingly relevant test failure.

peppy:
I'd be testing commands on a per-run basis (since they are all manual runs).

bdach:
> I'd be testing commands on a per-run basis (since they are all manual runs).

I'm assuming this means "I'll fix any issues with the other commands as I run into them", so I guess I have no real objections at that point.

PR #326, opened 2025-03-06
Remove envvar to disable global rank updates

peppy:
This would still update `rank_score`, which we're not doing going forward. So rather than have an explicit envvar, it's easier to just disable the whole `UserTotalPerformanceProcessor` during deploy stages.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserTotalPerformanceProcessor.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserTotalPerformanceProcessor.cs
@@ -30,8 +30,6 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Processors
         private readonly ConcurrentDictionary<int, MemoryCache> rankScoreIndexPartitionCache =
             new ConcurrentDictionary<int, MemoryCache>();
 
-        private static readonly bool update_global_ranks = Environment.GetEnvironmentVariable("UPDATE_GLOBAL_RANKS") != "0";
-
         public void RevertFromUserStats(SoloScore score, UserStats userStats, int previousVersion, MySqlConnection conn, MySqlTransaction transaction)
         {
         }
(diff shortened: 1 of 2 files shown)


osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserTotalPerformanceProcessor.cs
```diff
@@ -85,7 +83,7 @@ public async Task UpdateUserStatsAsync(UserStats userStats, int rulesetId, MySql
 
             (userStats.rank_score, userStats.accuracy_new) = UserTotalPerformanceAggregateHelper.CalculateUserTotalPerformanceAggregates(userStats.user_id, scores);
 
-            if (updateIndex)
+            if (updateMilestones)
```
bdach, line 86: At this point you could argue that instead of having a flag for this you just remove it and have the one usage that passes `updateMilestones: true` call `updateGlobalRank()` itself after calling this method and the other one that passes `false` - just not. Not sure which is better. I'd argue this is already twisty and confusing given `UpdateUserStatsAsync()` is xmldoc'd to "not insert new stats values into the database" but `updateGlobalRank()` *does* write to database, and making that change I propose would kinda fix that.

peppy: Sounds good, adjusted.

bdach (approved):
As long as there is some other process for doing a full sweep of the global / historical rank updates in place, and that commands are used for deploys exclusively, this makes sense to me

peppy:
For reference, that command is (still) https://github.com/peppy/osu-web-10/blob/master/www/web/update_pp_rank_history.php

bdach:
> For reference, that command is (still) https://github.com/peppy/osu-web-10/blob/master/www/web/update_pp_rank_history.php

One concern I'd have here is that that command doesn't appear to touch `osu_user_performance_rank_highest` table at all, which the `updateGlobalRank()` method here did update.

peppy:
> One concern I'd have here is that that command doesn't appear to touch osu_user_performance_rank_highest table at all, which the updateGlobalRank() method here did update.

Yeah it's assumed this will be updated the next time a user sets a score. Intended IMO.

Issue #324, opened 2025-03-01
Scoring attributes cache is not invalidated on disqualify / unrank


peppy:
https://github.com/ppy/osu-queue-score-statistics/blob/8f44b2d76feccbc2a9c6b51e357e67de2c39fd39/osu.Server.Queues.ScoreStatisticsProcessor/Helpers/BatchInserter.cs#L353-L354

https://discord.com/channels/188630481301012481/1097318920991559880/1345275732204716092

PR #327, opened 2025-03-17
Add invalidation flow to `BatchInserter`

peppy:
Closes https://github.com/ppy/osu-queue-score-statistics/issues/324

I went through multiple iterations, trying to make `BatchInserter` use `BeatmapStore` but it's all a bit of a PITA due to slightly different interfaces. So as a starting point, I've fixed the problematic case in isolation.

`BeatmapStore` can benefit from similar invalidation but requires a slightly different approach due to `MemoryCache` usage. It's also not as urgent because it currently purges itself [every minute](https://github.com/ppy/osu-queue-score-statistics/blob/0eaf633c6495aa37bcd421336def85df04faedae/osu.Server.Queues.ScoreStatisticsProcessor/Processors/ScorePerformanceProcessor.cs#L129-L135). I will PR a separate change for that later today or tomorrow.

Tested locally:

https://github.com/user-attachments/assets/93ff2463-afd0-4e01-8f2a-959e27df8f91

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Helpers/BatchInserter.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Helpers/BatchInserter.cs
@@ -59,6 +59,40 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Helpers
 
         public List<ScoreItem> ScoreStatisticsItems { get; } = new List<ScoreItem>();
 
+        static BatchInserter()
+        {
+            _ = BeatmapStatusWatcher.StartPollingAsync(updates =>
+            {
+                foreach (int beatmapSetId in updates.BeatmapSetIDs)
+                {
+                    using (var db = DatabaseAccess.GetConnection())
+                    {
+                        ICollection<BeatmapLookup> scoringAttributesKeys = scoring_attributes_cache.Keys;
+                        ICollection<DifficultyAttributesLookup> difficultyAttributesKeys = attributes_cache.Keys;
+
+                        foreach (int beatmapId in db.Query<int>("SELECT beatmap_id FROM osu_beatmaps WHERE beatmapset_id = @beatmapSetId AND `deleted_at` IS NULL",
+                                     new { beatmapSetId }))
+                        {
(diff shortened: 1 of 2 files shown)


smoogipoo (approved):
Seems to read fine to me

osu.Server.Queues.ScoreStatisticsProcessor/Helpers/BatchInserter.cs
```diff
@@ -377,7 +417,7 @@ private static LegacyBeatmapConversionDifficultyInfo getDifficultyInfo(int beatm
 
             using (var connection = DatabaseAccess.GetConnection())
             {
-                Beatmap beatmap = connection.QuerySingle<Beatmap>("SELECT * FROM osu_beatmaps WHERE `beatmap_id` = @BeatmapId", new
+                Beatmap beatmap = connection.QuerySingle<Beatmap>("SELECT * FROM osu_beatmaps WHERE `beatmap_id` = @BeatmapId ", new
```
bdach, line 420: weird space insertion here

osu.Server.Queues.ScoreStatisticsProcessor/Helpers/BatchInserter.cs
```diff
@@ -459,13 +499,6 @@ private async Task enqueueForFurtherProcessing(ulong firstId, ulong lastId, MySq
                 }
                 else
                 {
-                    if (dryRun)
-                    {
-                        // We can't retrieve this from the database because it hasn't been inserted.
-                        ScoreStatisticsItems.Add(new ScoreItem(new SoloScore(), new ProcessHistory()));
-                        return;
-                    }
```
bdach, line 467: what was this ever doing :thinking: seems just straight wrong in retrospect

bdach (approved):

osu.Server.Queues.ScoreStatisticsProcessor/Helpers/BatchInserter.cs
peppy: It was working before `BatchInserter` was a thing. Just never conveyed across properly (only blocks queue insertion before this PR).

PR #328, opened 2025-03-18
Remove store invalidation from score processing flow

peppy:
Tested the build store part locally for correctness.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Stores/BuildStore.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Stores/BuildStore.cs
@@ -1,10 +1,10 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
-using System.Collections.Generic;
-using System.Linq;
+using System;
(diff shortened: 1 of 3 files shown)


smoogipoo (approved):
Lgtm

osu.Server.Queues.ScoreStatisticsProcessor/Stores/BuildStore.cs
```diff
@@ -15,34 +15,27 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Stores
     /// </summary>
     public class BuildStore
     {
-        private readonly IReadOnlyDictionary<int, Build> builds;
+        private static readonly uint memory_cache_size_limit = uint.Parse(Environment.GetEnvironmentVariable("MEMORY_CACHE_SIZE_LIMIT") ?? "128000000");
```
bdach, line 18: Probably not gonna be an issue due to the sliding expiration & overall expected low number of items here but I sort of wonder about this, three memory caches disparately read `MEMORY_CACHE_SIZE_LIMIT` for their own size limit, so I'm pretty sure the limit is going to be per cache rather than per the full process, which is maybe fine but I don't know if this was the intention

bdach (approved):
let's see what happens i guess

PR #329, opened 2025-03-24
Encapsulate difficulty attribute lookup logic in medal processing code to prevent invalid lookups

bdach:
Since 2f603b7c22f6b53de53b4718394b96e8c562c053, difficulty attribute lookups no longer soft-fail if the rows are missing in the database, and instead hard-fail. In addition, from the latest star rating recalculation onward, only `approved > 0` beatmaps have difficulty attributes populated in the database.

This makes `MedalAwarderContext.BeatmapStore.GetDifficultyAttributesAsync()` unsafe, as medal processors (in particular, hush-hush ones) will attempt beatmap attribute lookups that may not necessarily hit *ranked* beatmaps, which will henceforth hard-fail rather than return `null` as previously.

To prevent this from happening, prevent medal awarders from being able to directly retrieve difficulty attributes via the beatmap store, and instead add a redirection method that does the `approved > 0` check, and does a soft fallback to `null` if it is called on an unranked beatmap.


peppy (approved):
Seems logical

PR #330, opened 2025-03-24
Make `BeatmapStore` static

peppy:
This fixes the fact that we were instantiating one of these per scores processed, causing thousands of background pollers 🙈.

It also means the caches are now shared over multiple usages, which is beneficial.


smoogipoo (approved):
Looks reasonable

bdach (approved):
let's see how this goes I guess

PR #331, opened 2025-03-28
Add support for mania variant ranked score bookkeeping and verification

peppy:
Requested by web team for https://github.com/ppy/osu-web/pull/12012.

Have tested to the best I can on production (verify command at least). Will run post-deploy so we don't have any non-tracked period.


smoogipoo:
Updated diffcalc_sheet_generator in https://github.com/smoogipoo/diffcalc-sheet-generator/pull/19

smoogipoo (approved):
Code-wise seems okay to me

PR #333, opened 2025-04-27
Add Hidden as a relevant legacy mod for osu ruleset

tsunyoku:
Required as a result of https://github.com/ppy/osu/pull/31351.

Above change is a difficulty related change in the `pp-dev` branch, so this PR can be ignored until the next deploy cycle (where it will become a requirement).

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor.Tests/LegacyModsHelperTests.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor.Tests/LegacyModsHelperTests.cs
@@ -11,15 +11,7 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Tests
     public class LegacyModsHelperTests
     {
         [Fact]
-        public void OsuWithoutFlashlight()
-        {
-            Assert.Equal(
-                LegacyMods.DoubleTime | LegacyMods.HalfTime | LegacyMods.HardRock | LegacyMods.Easy | LegacyMods.TouchDevice,
-                LegacyModsHelper.MaskRelevantMods(LegacyModsHelper.ALL_MODS & ~LegacyMods.Flashlight, false, 0));
-        }
-
-        [Fact]
-        public void OsuWithFlashlight()
+        public void Osu()
         {
             Assert.Equal(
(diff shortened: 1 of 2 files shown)


bdach:
This looks fine on its own, but the thing that I'm missing in all of this is that I don't see it mentioned anywhere that this seems to double-to-triple the required storage for beatmap attributes? Like, if Hidden and Spun Out now affect difficulty attributes, then as far as I can tell, this changes the number of combinations that will need to be stored from 54 to 144.

It's probably nowhere near mania levels, but my back-of-napkin math has this change increasing the storage size of beatmap attributes from ~0.8GB to ~2.1GB. And probably increases the duration of recalculations in the future, too. Were these implications discussed at any point in review of the original change?

ping @peppy for awareness

tsunyoku:
Yeah, that is true. In my head https://github.com/ppy/osu/pull/31735 gave me a lot of headroom since we already made a lot of reductions to the space used since that PR.

If this sort of increase can't be justified, I can probably rollback the SpunOut part specifically (it's less important) if that brings the amount of combinations to a more acceptable number.

smoogipoo:
Also if this change is to accepted, it should not go into `master`. We would need a similar `pp-dev` branch here.

I wonder if we should wait for https://github.com/ppy/osu-queue-score-statistics/pull/274 instead?

tsunyoku:
As far as I was told, there's no plans to merge #274 soon unless there is specifically good reason for it. That's a separate discussion that I plan to touch on in that PR, but waiting on it might just end up being a blocker for the change that relies on this.

Also, I'm unsure what #274 will solve for this - HD and SO are both legacy mods so even with partial real time it'd continue to go to database. If you want to change that, then that's a huge consequence to the original scope of that PR.

Do we really need a `pp-dev` branch here for this? It's not really an often case where changes to `osu-queue-score-statistics` is required for PP, so leaving the PR to sit until it's required seems fine to me but I'm indifferent about it either way.

tsunyoku:
As per https://github.com/ppy/osu/pull/34838, this now only requires adding Hidden rather than Hidden _and_ SpunOut.

PR #335, opened 2025-06-09
Fix mania keymode user stats processor completely bungling convert treatment

bdach:
The TODO about key mods is one thing, the flagrantly idiotic reinterpretation of `diff_size` as key count always regardless of original beatmap ruleset is another.

This is not a complete solution because the "initial playcount estimation" is pretty much infeasible to include converts in, but it is at least supposed to stop the bleeding on some complete idiocy.

And yeah this will likely also mean recomputation of these stats game wide. Probably.

I would ask who bungled the previous implementation *that* badly, but I know I would need to look in the mirror to answer that question.

---

The alternative solution here would be to delete converts from consideration via another guard, probably like so:

```diff
diff --git a/osu.Server.Queues.ScoreStatisticsProcessor/Processors/ManiaKeyModeUserStatsProcessor.cs b/osu.Server.Queues.ScoreStatisticsProcessor/Processors/ManiaKeyModeUserStatsProcessor.cs
index 160438e..3eedcd9 100644
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Processors/ManiaKeyModeUserStatsProcessor.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Processors/ManiaKeyModeUserStatsProcessor.cs
@@ -33,11 +33,10 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Processors
             if (score.ruleset_id != 3)
                 return;
 
-            if (score.beatmap == null)
+            if (score.beatmap == null || score.beatmap.playmode != 3)
                 return;
 
             int keyCount = (int)score.beatmap.diff_size;
-            // TODO: reconsider when handling key conversion mods in the future?
             if (keyCount != 4 && keyCount != 7)
                 return;
 

```

but I dunno. That feels worse.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor.Tests/ManiaKeyModeUserStatsProcessorTests.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor.Tests/ManiaKeyModeUserStatsProcessorTests.cs
@@ -4,6 +4,8 @@
 using System;
 using System.Text;
 using Dapper;
+using osu.Game.Online.API;
+using osu.Game.Rulesets.Mania.Mods;
 using osu.Game.Scoring;
 using Xunit;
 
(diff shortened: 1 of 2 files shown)


smoogipoo (approved):
From a code-only perspective this looks correct to me. I also wouldn't mind the alternative solution proposed in the OP.

peppy (approved):

PR #337, opened 2025-07-24
Emit lazer leaderboard rank events when processing lazer scores

bdach:
- Would close https://github.com/ppy/osu-web/issues/11188
- [x] Depends on database migration from https://github.com/ppy/osu-web/pull/12303

PRing as draft because this is doing Weird Things (API calls to web to fetch the score's rank) because of the direction I was given internally and I want to make sure everyone is on board with the full implications of that.

The diff is a bit large because I got weirded out by the legacy interop request logic being in `DatabaseHelper` and I wanted to conjure a `WebHelper` anyway for handling of API.

I've tested the happy path once but not much else. Probably more testing around the token expiry logic is warranted, but I'll do that once I'm confident this diff doesn't get dumpstered on entry.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Helpers/WebRequestHelper.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Helpers/WebRequestHelper.cs
@@ -1,0 +1,109 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using System.Linq;
+using System.Net;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Security.Cryptography;
+using System.Text;
+using System.Threading;
+using Newtonsoft.Json;
+
+namespace osu.Server.Queues.ScoreStatisticsProcessor.Helpers
+{
+    public static class WebRequestHelper
+    {
+        private static readonly string? shared_interop_domain = Environment.GetEnvironmentVariable("SHARED_INTEROP_DOMAIN");
+        private static readonly string? shared_interop_secret = Environment.GetEnvironmentVariable("SHARED_INTEROP_SECRET");
+
+        private static readonly HttpClient http = new HttpClient();
+
+        public static HttpResponseMessage RunSharedInteropCommand(string command, string method = "GET", dynamic? postObject = null)
+        {
+            if (string.IsNullOrEmpty(shared_interop_domain))
+            {
+#if !DEBUG
+                throw new InvalidOperationException($"Attempted legacy IO call without target domain specified ({command})");
+#endif
+                return null!;
+            }
+
+            if (string.IsNullOrEmpty(shared_interop_secret))
+            {
+#if !DEBUG
+                throw new InvalidOperationException($"Attempted legacy IO call without secret set ({command}");
+#endif
+                return null!;
+            }
+
+            int retryCount = 3;
+
+            retry:
+
+            try
+            {
+                long time = DateTimeOffset.UtcNow.ToUnixTimeSeconds();
+
+                string url = $"{shared_interop_domain}/_lio/{command}{(command.Contains('?') ? "&" : "?")}timestamp={time}";
+                string signature = hmacEncode(url, Encoding.UTF8.GetBytes(shared_interop_secret));
+
+#pragma warning disable SYSLIB0014
+                var request = WebRequest.CreateHttp(url);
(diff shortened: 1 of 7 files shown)


osu.Server.Queues.ScoreStatisticsProcessor/Helpers/WebRequestHelper.cs
```diff
@@ -0,0 +1,222 @@
...
+            if (accessToken != null && accessTokenExpiry != null && DateTimeOffset.Now <= accessTokenExpiry.Value.AddMinutes(-5))
+                return accessToken;
+
+            var requestMsg = new HttpRequestMessage(HttpMethod.Post, $"{shared_interop_domain}/oauth/token");
+            requestMsg.Content = new FormUrlEncodedContent([
+                new KeyValuePair<string, string>("client_id", api_client_id),
+                new KeyValuePair<string, string>("client_secret", api_client_secret),
+                new KeyValuePair<string, string>("grant_type", "client_credentials"),
+                new KeyValuePair<string, string>("scope", "public"),
+            ]);
```
bdach, line 193: This sets up the authentication with the API as stated in the OP.

The lazer API client id/secret pair cannot be reused here for two reasons: that token only supports `*` scope (`public` scope [requires the client to have an user owning it](https://github.com/ppy/osu-web/blob/904c94936f2e4b87c88c2dbfd37e74b2b107fca1/app/Models/OAuth/Token.php#L183-L186)), and [the `*` scope doesn't support `client_credentials` grant](https://github.com/ppy/osu-web/blob/904c94936f2e4b87c88c2dbfd37e74b2b107fca1/app/Models/OAuth/Token.php#L189-L191).

So this needs to either be a user-owned client (by whom?) or I'm going to need @ppy/team-web to carve me out something here.

osu.Server.Queues.ScoreStatisticsProcessor/Helpers/WebRequestHelper.cs
```diff
@@ -0,0 +1,222 @@
...
+        /// At this point there are two options: either hit ES directly from here, or hit <c>osu-web</c> which already does the same trick to retrieve global rank of a score in the beatmap leaderboard.
+        /// The first approach was deemed undesirable, because it is preferred that there is a single connection point to ES in the infrastructure.
+        /// Therefore, the <c>GET /scores/{id}</c> API endpoint should do the trick for the second approach.
+        /// If this ever becomes a problem the proposition is to move the ES lookup logic to https://github.com/ppy/osu-global-rank-lookup-cache and have both web and this call into that component.
+        /// </para>
+        /// <para>
+        /// Note that this working correctly also relies on the fact that <c>osu-web</c> API has lazer mode turned forcefully on with no way to toggle it off
+        /// (see https://github.com/ppy/osu-web/blob/f46806bb81eb0d3b0807c70bf8e3dc13f0783ec9/app/Libraries/Search/ScoreSearchParams.php#L86-L88).
+        /// </para>
+        /// </remarks>
```
bdach, line 145: I'd like anyone reviewing this to read *all of this* over *carefully* and double-check all of the claims made here because this is a pretty precarious setup.

osu.Server.Queues.ScoreStatisticsProcessor/Helpers/WebRequestHelper.cs
```diff
@@ -0,0 +1,222 @@
...
+                return null;
+            }
+
+            accessToken = tokenResponse.access_token;
+            accessTokenExpiry = DateTimeOffset.Now.AddSeconds(tokenResponse.expires_in);
+            return accessToken;
+        }
+
+        // ReSharper disable All
+        private class TokenResponse
```
peppy, line 211: Can probably use `osu.Game.Online.API.OAuthToken`. And arguably, the whole `OAuth` class, which would also mean you'd be using refresh tokens correctly?

osu.Server.Queues.ScoreStatisticsProcessor/Helpers/WebRequestHelper.cs
peppy: This all reads correctly to me. Reading over this, the only real concern I would have is with web API not being able to handle the number of requests it receives, but since there's no large JSON decode involved and it's being limited to user high scores already, I think we should be okay.

osu.Server.Queues.ScoreStatisticsProcessor/Helpers/WebRequestHelper.cs
peppy: Right.

peppy (approved):

PR #344, opened 2025-10-21
Write last processed score id to counts table

peppy:
For use in `osu-web` for knowing whether a scores has been processed or not, now that it is not guaranteed that `score_process_history` rows exist for eternity.

Note that this may be slightly inaccurate if there are multiple runners processing the queue, but only by sub-second values.


peppy:
I did mention in the commit, but probably good to have an inline comment for future visitors too.

bdach:
It's fine without.

PR #369, opened 2026-05-20
Disallow all processors from running if score is too short

peppy:
Have not run tests locally yet, just a quick PoC based on [recent report](https://discord.com/channels/188630481301012481/188630652340404224/1506626367264133151).

FWIW I think we should guard against this client side as well (client should not submit under said circumstances eagerly, to reduce network / client strain.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/ScoreStatisticsQueueProcessor.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/ScoreStatisticsQueueProcessor.cs
@@ -204,6 +204,9 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor
                         BeatmapSetId = score.beatmap.beatmapset_id
                     });
 
+                    if (!score.IsValidForPlayTracking())
+                        return;
+
                     using (var transaction = conn.BeginTransaction(IsolationLevel.ReadCommitted))
                     {
                         var userStats = DatabaseHelper.GetUserStatsAsync(score, conn, transaction).Result;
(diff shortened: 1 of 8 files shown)


smoogipoo (approved):

peppy:
Keep in mind some test fixes here are required. Fixing today.

peppy:
I've fixed things up but had to change how this is working, please see individual commits for commentary.

osu.Server.Queues.ScoreStatisticsProcessor/ScoreStatisticsQueueProcessor.cs
```diff
@@ -216,44 +216,47 @@ private void processScore(ScoreItem item, bool force = false)
                             return;
                         }
 
-                        // if required, we can rollback any previous version of processing then reapply with the latest.
-                        if (item.ProcessHistory != null)
+                        if (score.IsValidForPlayTracking())
```
bdach, line 219: - With this implementation, even processors that specify `RunOnFailedScores => true` may not ever see a score if it fails to clear the hit count / play duration criteria outlined by this method. This includes: hit statistics, last played date, medals, total user play count, total user play time, total user score. Please confirm that all of those outcomes are intended.
- This is the only special path in the processing loop that does not attempt to add a datadog tag to the score in some way to denote why the score's processing was skipped.

peppy: > This includes: hit statistics, last played date, medals, total user play count, total user play time, total user score. Please confirm that all of those outcomes are intended.

That's kind of the whole point of this PR, so yeah, it's intentional. The only one you could maybe argue against is last played, but at the point you're playing so short that it doesn't register as a score, is it really considered playing?

> This is the only special path in the processing loop that does not attempt to add a datadog tag to the score in some way to denote why the score's processing was skipped.

I'll add a tag.

bdach (approved):

PR #370, opened 2026-05-21
Account for legacy score imports which do not have `started_at` populated

peppy:


osu.Server.Queues.ScoreStatisticsProcessor/Helpers/PlayValidityHelper.cs
```diff
@@ -26,6 +26,12 @@ public static bool IsValidForPlayTracking(this SoloScore score)
             if (score.passed)
                 return true;
 
+            // `started_at` is not populated for legacy scores...
+            // but legacy scores are only imported into the new table if they have already passed length validity checks so we can bypass this check.
+            // see https://github.com/peppy/osu-web-10/blob/6031aed7a7382b2b3b44cedde6c2f8a19f0bf378/www/web/osu-submit-20190809.php#L234-L235
+            if (score.is_legacy_score && score.started_at == null)
```
peppy: Second condition can be dropped if preferred. It's guaranteed.

osu.Server.Queues.ScoreStatisticsProcessor/Helpers/PlayValidityHelper.cs
```diff
@@ -26,6 +26,12 @@ public static bool IsValidForPlayTracking(this SoloScore score)
             if (score.passed)
                 return true;
 
+            // `started_at` is not populated for legacy scores...
+            // but legacy scores are only imported into the new table if they have already passed length validity checks so we can bypass this check.
+            // see https://github.com/peppy/osu-web-10/blob/6031aed7a7382b2b3b44cedde6c2f8a19f0bf378/www/web/osu-submit-20190809.php#L234-L235
```
bdach, line 31: This has the caveat that the web-10 checks mentioned here do not match the ones implemented by this method one-to-one, as they check *only* the duration.

peppy: 🤫

peppy: can split the conditions out i guess? and just skip the duration one

bdach (approved):

peppy:
Double-check [4a3ae74](https://github.com/ppy/osu-queue-score-statistics/pull/370/commits/4a3ae741aa3d86ef060a6dbdeec8e8d4fb5f911b) please 🙏

bdach (approved):

PR #374, opened 2026-05-29
Adjust mod multiplier recalculation processes to minimise number of required database writes

bdach:
- [x] Depends on the actual game-side mod multiplier changes
- [x] Will want to recheck further myself with a fresh brain

PRing early as draft primarily for visibility reasons.

The original plan for deploying the mod multiplier changes was to first ensure that all scores in the database have `total_score_without_mods` present, and then write out the updated `total_score` across the whole table as the new multipliers times that populated `total_score_without_mods` quantity.

<details>
<summary>Aside: Why is the migration not run in-place?</summary>

The primary reason for me approaching things in this way rather than running the recalculation in-place is because running the recalculation in-place would be an inherently *lossy* process.

- Assuming everything goes right, because total score is rounded to full integers and mod multipliers are very much not integers, the possibility of rounding errors and floating-point inaccuracies is very real. Therefore, any in-place recalculation of multipliers would be a process that ideally would *only ever occur once and never again*, because the more times it is performed, the bigger the degradation incurred with every in-place conversion.
- And that is *assuming everything goes right*. In case of *any* mistake or incorrectly-applied migration, the original data would be *lost* in an in-place migration, and would not be recoverable quickly or at all because of how complicated the scoring algorithm is at the explicit demand of the user base. The closest thing to a recovery procedure in that case would be replays - *when there are replays* - and even that would be slow.

  It is my former professional experience that tells me you **never** want to run huge data migrations in-place which incur chances of irretrievable or difficult-to-recover-from data loss.

</details>

In recent days, an attempt was made to run the backpopulation of `total_score_without_mods` on the producton `scores` table. Unfortunately at this time the original plan appears to be mostly unviable.

The number of rows in the production `scores` table that require this backfill is estimated to be around 2.38 billion rows. This would incur a storage overhead in the hundreds of gigabytes (not only from the actual rows themselves, but also from a transient spike in the size of binlog, which is relevant when replication is depended upon) and take roughly a month to execute.

This PR contains a set of changes that is designed to reduce the number of writes required to execute this migration. The cost here is the increased complexity of the migration process. While I am *somewhat* confident this is going to work, I will want to test this more thoroughly on the actual changes and maybe even do some test runs locally using data dumps to make sure I have not missed anything.

Additionally, comments from https://github.com/ppy/osu-queue-score-statistics/pull/269#pullrequestreview-4370002192 are addressed here.

As for the gory details:

## [Switch to new mod multiplier calculation API in backpopulation command](https://github.com/ppy/osu-queue-score-statistics/commit/ebd58a635aa5c28a55d436f9f231919116c4701a)

By itself this does nothing, the old API worked *fine* as far as the backpopulation was concerned. However doing this allows for an opportunity of automatically handling the change to mania key mod multipliers that was done live without backwards adjustments. More on this later.

## [Adjust query in backpopulation command to target the minimum necessary number of rows](https://github.com/ppy/osu-queue-score-statistics/commit/a5bc17d5e1dd36ac070c25e1c36f5507da6d8d83)

As the inline commentary states, the number of rows that require writes of `total_score_without_mods` can be significantly reduced because of two facts:

- Firstly, when it comes to scores set on stable, `total_score_without_mods` as well as `total_score` are quantities derived directly from `legacy_total_score` as well as beatmap attributes. Therefore, it is not actually necessary to write `total_score_without_mods` for those scores, and instead the score conversion algorithm can be used to recalculate the multipliers instead.
- Secondly, when it comes to lazer scores specifically, this change assumes the invariant that if there are no mods on a score, then `total_score == total_score_without_mods` - so there is no point writing out values.

## [Adjust backpopulation command to automatically handle change to mania key mod multipliers](https://github.com/ppy/osu-queue-score-statistics/commit/28e2933bb2550d122cf87d2087753a81b5381b84)

Migrating to the new multiplier API, which already has to handle multiplier versioning and the directly-applied change to mania key mod multipliers, means that it is now possible to leverage the versioned logic and automatically correct for the key mod multiplier change by filling out `ScoreInfo.ClientVersion` using `osu_builds`.

## [Adjust mod multiplier recalculation command to work with the minimum necessary number of rows with `total_score_without_mods`](https://github.com/ppy/osu-queue-score-statistics/commit/1186659a81986461fbb91dcbd64dd25bbb5bc9a2)

Adds an alternate processing path for stable scores which leverages the algorithm of converting from legacy total score to implement the multiplier recalculation.

The command was already skipping scores with no mods at all, so no adjustment required there.

## [Add test coverage of commands](https://github.com/ppy/osu-queue-score-statistics/commit/e0b8c101e2379ebd1f0f652ddc7fcfbb14eea69c)

Written using real-world cases pulled from live production scores (via data.ppy.sh dumps plus some manual trawling for the mania key mod cases).

In the current state the most useful tests are the ones checking correctness of the mania key mod recalculation. Everything else will need updating once the new proposed values are out.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/PopulateTotalScoreWithoutModsCommand.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/PopulateTotalScoreWithoutModsCommand.cs
@@ -110,8 +110,12 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Maintenance
 
                     score.beatmap = beatmap;
                     var scoreInfo = score.ToScoreInfo();
-                    if (score.build_id != null && buildsById.TryGetValue(score.build_id.Value, out var build))
-                        scoreInfo.ClientVersion = build.version;
+
+                    if (score.build_id == null || !buildsById.TryGetValue(score.build_id.Value, out Build? build))
+                        throw new InvalidOperationException($"Aborting: score {score.id} has missing or invalid build ID of {score.build_id}!");
+
+                    scoreInfo.ClientVersion = build.version;
+
                     var ruleset = LegacyRulesetHelper.GetRulesetFromLegacyId(scoreInfo.RulesetID);
 
                     var scoreMultiplierCalculator = ruleset.CreateScoreMultiplierCalculator(new ScoreMultiplierContext(beatmap.GetLegacyBeatmapConversionDifficultyInfo(), scoreInfo));
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/PopulateTotalScoreWithoutModsCommand.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/PopulateTotalScoreWithoutModsCommand.cs
@@ -50,6 +50,13 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Maintenance
 
             await Task.Delay(5000, cancellationToken);
 
+            // prefetch a mapping of lazer / tachyon build IDs to string versions for later use.
+            // note that `osu_builds` has two kinds of rows for builds: "main" rows (with `stream_id` set),
+            // and "platform-specific" rows (with platform suffixes in `version`, as well as null `stream_id`).
+            // `scores.build_id` points at the *platform-specific* rows, so filtering by stream ID cannot be easily done here.
+            var lazerBuildVersionsById = (await conn.QueryAsync<Build>(@"SELECT `build_id`, `version` FROM `osu_builds`"))
+                .ToDictionary(build => build.build_id, build => build.version);
+
             while (!cancellationToken.IsCancellationRequested)
             {
                 var scoresToPopulate = (await conn.QueryAsync<SoloScore>(
@@ -96,10 +103,6 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Maintenance
                 var beatmapsById = (await conn.QueryAsync<Beatmap>(@"SELECT * FROM `osu_beatmaps` WHERE `beatmap_id` IN @ids", new { ids = beatmapIds }))
                     .ToDictionary(beatmap => beatmap.beatmap_id);
 
-                var buildIds = scoresToPopulate.Select(score => score.build_id).Where(id => id != null).Cast<ushort>().ToHashSet();
-                var buildsById = (await conn.QueryAsync<Build>(@"SELECT * FROM `osu_builds` WHERE `build_id` IN @ids", new { ids = buildIds }))
-                    .ToDictionary(build => build.build_id);
-
                 foreach (var score in scoresToPopulate)
                 {
                     if (!beatmapsById.TryGetValue(score.beatmap_id, out var beatmap))
@@ -111,10 +114,10 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Maintenance
                     score.beatmap = beatmap;
                     var scoreInfo = score.ToScoreInfo();
 
-                    if (score.build_id == null || !buildsById.TryGetValue(score.build_id.Value, out Build? build))
+                    if (score.build_id == null || !lazerBuildVersionsById.TryGetValue(score.build_id.Value, out string? buildVersion))
                         throw new InvalidOperationException($"Aborting: score {score.id} has missing or invalid build ID of {score.build_id}!");
 
-                    scoreInfo.ClientVersion = build.version;
+                    scoreInfo.ClientVersion = buildVersion;
 
                     var ruleset = LegacyRulesetHelper.GetRulesetFromLegacyId(scoreInfo.RulesetID);
 
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/PopulateTotalScoreWithoutModsCommand.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/PopulateTotalScoreWithoutModsCommand.cs
@@ -92,9 +92,13 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Maintenance
                     continue;
                 }
 
-                uint[] beatmapIds = scoresToPopulate.Select(score => score.beatmap_id).Distinct().ToArray();
+                var beatmapIds = scoresToPopulate.Select(score => score.beatmap_id).ToHashSet();
                 var beatmapsById = (await conn.QueryAsync<Beatmap>(@"SELECT * FROM `osu_beatmaps` WHERE `beatmap_id` IN @ids", new { ids = beatmapIds }))
                     .ToDictionary(beatmap => beatmap.beatmap_id);
+
+                var buildIds = scoresToPopulate.Select(score => score.build_id).Where(id => id != null).Cast<ushort>().ToHashSet();
+                var buildsById = (await conn.QueryAsync<Build>(@"SELECT * FROM `osu_builds` WHERE `build_id` IN @ids", new { ids = buildIds }))
+                    .ToDictionary(build => build.build_id);
 
                 foreach (var score in scoresToPopulate)
                 {
@@ -106,6 +110,8 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Maintenance
 
                     score.beatmap = beatmap;
                     var scoreInfo = score.ToScoreInfo();
+                    if (score.build_id != null && buildsById.TryGetValue(score.build_id.Value, out var build))
+                        scoreInfo.ClientVersion = build.version;
                     var ruleset = LegacyRulesetHelper.GetRulesetFromLegacyId(scoreInfo.RulesetID);
 
                     var scoreMultiplierCalculator = ruleset.CreateScoreMultiplierCalculator(new ScoreMultiplierContext(beatmap.GetLegacyBeatmapConversionDifficultyInfo(), scoreInfo));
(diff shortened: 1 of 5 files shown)


osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/PopulateTotalScoreWithoutModsCommand.cs
```diff
@@ -92,10 +92,14 @@ ORDER BY `id`
                     continue;
                 }
 
-                uint[] beatmapIds = scoresToPopulate.Select(score => score.beatmap_id).Distinct().ToArray();
+                var beatmapIds = scoresToPopulate.Select(score => score.beatmap_id).ToHashSet();
                 var beatmapsById = (await conn.QueryAsync<Beatmap>(@"SELECT * FROM `osu_beatmaps` WHERE `beatmap_id` IN @ids", new { ids = beatmapIds }))
                     .ToDictionary(beatmap => beatmap.beatmap_id);
 
+                var buildIds = scoresToPopulate.Select(score => score.build_id).Where(id => id != null).Cast<ushort>().ToHashSet();
```
peppy: Could potentially move this to a once-off fetch of all valid/relevant builds before the batching part, but probably not a huge saving at the end of the day. The overheads are mostly going to be in the DB writes still.

```
root@db-master-newer:osu> select count(*) from osu_builds where stream_id in (1, 3, 5, 6, 7, 14, 17);
| count(*) |
|----------|
|     4081 |

1 row in set
Time: 0.127s
```

If not moving this out, it might make sense to limit fetching to just the version string, to reduce data overheads from the rest of the columns.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/PopulateTotalScoreWithoutModsCommand.cs
```diff
@@ -89,7 +110,13 @@ public async Task<int> OnExecuteAsync(CancellationToken cancellationToken)
 
                     score.beatmap = beatmap;
                     var scoreInfo = score.ToScoreInfo();
-                    LegacyScoreDecoder.PopulateTotalScoreWithoutMods(scoreInfo);
+                    if (score.build_id != null && buildsById.TryGetValue(score.build_id.Value, out var build))
+                        scoreInfo.ClientVersion = build.version;
+                    var ruleset = LegacyRulesetHelper.GetRulesetFromLegacyId(scoreInfo.RulesetID);
+
+                    var scoreMultiplierCalculator = ruleset.CreateScoreMultiplierCalculator(new ScoreMultiplierContext(beatmap.GetLegacyBeatmapConversionDifficultyInfo(), scoreInfo));
```
peppy, line 124: Can we do a package bump to get this into a compiling state? Or is there valid reason to hold off for the final implementation?

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/PopulateTotalScoreWithoutModsCommand.cs
```diff
@@ -51,19 +52,35 @@ public async Task<int> OnExecuteAsync(CancellationToken cancellationToken)
...
                     // if `data` does not have `total_score_without_mods` set at all, `data->'$.total_score_without_mods'` will return the typical SQL NULL,
                     // but if `data` has `{'total_score_without_mods': null}`, then `data->'$.total_score_without_mods'` will return `CAST('null' AS JSON)` which IS NOT NULL.
                     // We want to be matching both of these cases. Thus, we use `JSON_VALUE`, which bypasses this footgun.
-                    "SELECT * FROM scores WHERE `id` BETWEEN @lastId AND (@lastId + @batchSize - 1) AND JSON_VALUE(`data`, '$.total_score_without_mods') IS NULL ORDER BY `id`",
+                    """
+                    SELECT * FROM scores
+                    WHERE `id` BETWEEN @lastId AND (@lastId + @batchSize - 1)
+                        AND `legacy_score_id` IS NULL
+                        AND JSON_LENGTH(`data`, '$.mods') > 0
+                        AND JSON_VALUE(`data`, '$.total_score_without_mods') IS NULL
```
peppy, line 81: ✅

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/PopulateTotalScoreWithoutModsCommand.cs
```diff
@@ -106,6 +110,8 @@ ORDER BY `id`
 
                     score.beatmap = beatmap;
                     var scoreInfo = score.ToScoreInfo();
+                    if (score.build_id != null && buildsById.TryGetValue(score.build_id.Value, out var build))
```
peppy: Do we want some kind of assertion that this ends up in a populated state? I *think* we should be able to guarantee that.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/RecalculateModMultipliersCommand.cs
```diff
@@ -117,6 +152,18 @@ public async Task<int> OnExecuteAsync(CancellationToken cancellationToken)
             return 0;
         }
 
+        private static BeatmapScoringAttributes? getScoringAttributesFor(SoloScore score, MySqlConnection conn)
```
peppy, line 155: I wonder if we want to expose and use the cached path for this.

https://github.com/ppy/osu-queue-score-statistics/blob/1421bade6c0d5c7617eb85b480a3aec0f2af774b/osu.Server.Queues.ScoreStatisticsProcessor/Helpers/BatchInserter.cs#L348-L352

peppy (commented):
No major issues found.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/PopulateTotalScoreWithoutModsCommand.cs
bdach: A package bump can be done but it will need to be done again anyway once the new multipliers land in client to make the recalculation command work.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/RecalculateModMultipliersCommand.cs
bdach: I considered it. My primary concern was running out of memory because the aforementioned static dictionary is never cleared other than via `BeatmapStatusWatcher.StartPollingAsync()` which is only relevant when the actual data changes.

I'm not even sure why this seemingly doesn't explode in `BatchInserter`, really.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/PopulateTotalScoreWithoutModsCommand.cs
bdach: > I _think_ we should be able to guarantee that.

Theoretically yes, because the main case of `build_id == null` is legacy imported scores and those are already skipped by this command. In practice, I don't know. I've seen enough cases of anomalies in data at rest to be skeptical about soft invariants like that being upheld.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/RecalculateModMultipliersCommand.cs
peppy: There's a finite number of ranked beatmaps in the mentioned case (`BatchInserter` should only be dealing with ranked beatmap scores). It can most definitely fit in memory.

For our purposes here, I think we also consider unranked beatmaps, which may tip the scales.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/PopulateTotalScoreWithoutModsCommand.cs
bdach: Attempted in https://github.com/ppy/osu-queue-score-statistics/pull/374/commits/3546b027a4ff3aab984ddbf378335838ca26a833.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/PopulateTotalScoreWithoutModsCommand.cs
bdach: Attempted in https://github.com/ppy/osu-queue-score-statistics/pull/374/commits/6676129e524982d005a8271491ef217edf79cb2c.

Note that as per inline commentary there, constraining to particular `stream_id` values cannot be easily done.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/PopulateTotalScoreWithoutModsCommand.cs
bdach: I am not sure whether that was for this change specifically but there are NuGet packages from today that at least make this compile, so applied in https://github.com/ppy/osu-queue-score-statistics/pull/374/commits/942ac75a84b141ab1c0e29df6ff127df51d5d870.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/RecalculateModMultipliersCommand.cs
bdach: > For our purposes here, I think we also consider unranked beatmaps

There are presumably lazer scores set on unranked beatmaps, yes. I am not aware of any process culling them at this time.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/PopulateTotalScoreWithoutModsCommand.cs
peppy: that was the intention 👍

peppy (approved):

bdach:
I'll undraft and check off the boxes from the OP now, given that the indicated plan was to dry run this and check whether this new adjusted process is preserving existing multipliers (with the one exception of mania key mods).

Still bit scared of all this but testing added in https://github.com/ppy/osu-queue-score-statistics/pull/374/commits/9b87267a9be1c27b357c58d73dd34a2901a9356f should make sure I haven't screwed up dry run mode at least.

Will re-test once more with new multiplier values.

peppy (approved):

PR #377, opened 2026-06-04
Improve speed of score recalculation by caching per-score lookups

peppy:
Local testing shows ~800x improvement on dry run, but I'm also testing with production database at high latency so providing accurate benchmark results here is omitted (local db mirror is still nor reinit yet).

Fully aware of potential memory concerns for `GetCachedScoringAttributes`. Will deal with that if we need to (convert to `MemoryCache` like the more mature methods), but the actual amount of data stored is quite small, so I'm not sure it will be required.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/RecalculateModMultipliersCommand.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/RecalculateModMultipliersCommand.cs
@@ -31,10 +31,10 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Maintenance
         public int BatchSize { get; set; } = 5000;
 
         [Option(CommandOptionType.SingleOrNoValue, Template = "--dry-run")]
+        [MemberNotNullWhen(false, nameof(elasticQueuePusher))]
         public bool DryRun { get; set; }
 
         [Option(CommandOptionType.SingleOrNoValue, Template = "-v|--verbose", Description = "Output verbose information on processing.")]
-        [MemberNotNullWhen(false, nameof(elasticQueuePusher))]
         public bool Verbose { get; set; }
 
         private readonly StringBuilder sqlBuffer = new StringBuilder();
@@ -188,7 +188,7 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Maintenance
 
                     if (elasticItems.Count > 0)
                     {
-                        elasticQueuePusher!.PushToQueue(elasticItems.ToList());
+                        elasticQueuePusher.PushToQueue(elasticItems.ToList());
                         Console.WriteLine($"Queued {elasticItems.Count} items for indexing");
                     }
                 }
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/RecalculateModMultipliersCommand.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/RecalculateModMultipliersCommand.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Text;
 using System.Threading;
@@ -32,11 +33,13 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Maintenance
         public bool DryRun { get; set; }
 
         [Option(CommandOptionType.SingleOrNoValue, Template = "-v|--verbose", Description = "Output verbose information on processing.")]
+        [MemberNotNullWhen(false, nameof(elasticQueuePusher))]
         public bool Verbose { get; set; }
 
         private readonly StringBuilder sqlBuffer = new StringBuilder();
 
-        private readonly ElasticQueuePusher elasticQueuePusher = new ElasticQueuePusher();
+        private ElasticQueuePusher? elasticQueuePusher;
+
         private readonly List<ElasticQueuePusher.ElasticScoreItem> elasticItems = new List<ElasticQueuePusher.ElasticScoreItem>();
 
         [UsedImplicitly]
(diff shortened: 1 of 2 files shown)


osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/RecalculateModMultipliersCommand.cs
```diff
@@ -80,15 +90,13 @@ public async Task<int> OnExecuteAsync(CancellationToken cancellationToken)
...
-                uint[] beatmapIds = scoresWithMods.Select(score => score.beatmap_id).Distinct().ToArray();
-                var beatmapsById = (await conn.QueryAsync<Beatmap>(@"SELECT * FROM `osu_beatmaps` WHERE `beatmap_id` IN @ids", new { ids = beatmapIds }))
-                    .ToDictionary(beatmap => beatmap.beatmap_id);
-
                 foreach (var score in scoresWithMods)
                 {
                     string source = score.is_legacy_score ? "stable" : "lazer ";
 
-                    if (!beatmapsById.TryGetValue(score.beatmap_id, out var beatmap))
+                    var beatmap = await BeatmapStore.GetBeatmapAsync(score.beatmap_id, conn);
```
peppy, line 95: This is marginally slower to begin with due to less batching, but should even out over time.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/RecalculateModMultipliersCommand.cs
```diff
@@ -32,13 +34,17 @@ public class RecalculateModMultipliersCommand
         public bool DryRun { get; set; }
 
         [Option(CommandOptionType.SingleOrNoValue, Template = "-v|--verbose", Description = "Output verbose information on processing.")]
+        [MemberNotNullWhen(false, nameof(elasticQueuePusher))]
```
bdach: Nullability hint appears to be applied in the incorrect location. Existence of `elasticQueuePusher` is dependent on `DryRun` flag and not on `Verbose`.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/RecalculateModMultipliersCommand.cs
```diff
@@ -192,7 +188,7 @@ private void flush(MySqlConnection conn, bool force = false)
 
                     if (elasticItems.Count > 0)
                     {
-                        elasticQueuePusher.PushToQueue(elasticItems.ToList());
+                        elasticQueuePusher!.PushToQueue(elasticItems.ToList());
```
bdach: If above review comment is fixed this null-ignoring operator is no longer required.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/RecalculateModMultipliersCommand.cs
```diff
@@ -32,13 +34,17 @@ public class RecalculateModMultipliersCommand
...
         public bool Verbose { get; set; }
 
         private readonly StringBuilder sqlBuffer = new StringBuilder();
 
-        private readonly ElasticQueuePusher elasticQueuePusher = new ElasticQueuePusher();
+        private ElasticQueuePusher? elasticQueuePusher;
+
         private readonly List<ElasticQueuePusher.ElasticScoreItem> elasticItems = new List<ElasticQueuePusher.ElasticScoreItem>();
 
+        private Dictionary<uint, Beatmap> beatmapsById = new Dictionary<uint, Beatmap>();
```
bdach: Unused.

peppy:
Some stats:

this branch (running local for 15 minutes with 300 ms latency to database): 2.68m scores processed
live (running on production for 18 hours with 1 ms latency to database): 55.8m scores processed

should be a considerable improvement.

osu.Server.Queues.ScoreStatisticsProcessor/Helpers/BatchInserter.cs
```diff
@@ -348,22 +348,25 @@ public static ScoreInfo CreateReferenceScore(int rulesetId, HighScore highScore)
         private static readonly ConcurrentDictionary<BeatmapLookup, BeatmapScoringAttributes?> scoring_attributes_cache =
             new ConcurrentDictionary<BeatmapLookup, BeatmapScoringAttributes?>();
 
-        private static BeatmapScoringAttributes? getScoringAttributes(BeatmapLookup lookup)
+        public static BeatmapScoringAttributes? GetCachedScoringAttributes(BeatmapLookup lookup)
+        {
+            using (var conn = DatabaseAccess.GetConnection())
+                return GetCachedScoringAttributes(lookup, conn);
```
bdach: One concern with this rewrite is that the existing caller, namely `BatchInserter` itself, will now always take out a database connection even on a cache hit.

I'd be fine with copy-pasting the short-circuit path in here too to prevent that.

bdach (changes requested):

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/RecalculateModMultipliersCommand.cs
peppy: I copied this from another case and wondered why it never worked.

osu.Server.Queues.ScoreStatisticsProcessor/Helpers/BatchInserter.cs
peppy: good catch

osu.Server.Queues.ScoreStatisticsProcessor/Helpers/BatchInserter.cs
```diff
@@ -366,6 +366,21 @@ public static ScoreInfo CreateReferenceScore(int rulesetId, HighScore highScore)
...
+
+            BeatmapScoringAttributes? scoreAttributes = conn.QuerySingleOrDefault<BeatmapScoringAttributes>(
+                "SELECT * FROM osu_beatmap_scoring_attribs WHERE beatmap_id = @BeatmapId AND mode = @RulesetId", new
+                {
+                    BeatmapId = lookup.BeatmapId,
+                    RulesetId = lookup.RulesetId,
+                });
+
+            return scoring_attributes_cache[lookup] = scoreAttributes;
+        }
```
bdach, line 382: Not quite what I had in mind because this duplicates the actual logic of the retrieval, while I was suggesting only duplicating the guard, but sure I suppose.

bdach (approved):

osu.Server.Queues.ScoreStatisticsProcessor/Helpers/BatchInserter.cs
peppy: Saves doing two dictionary lookups 🤷

bdach: The flipside is needing to remember to touch both methods if one of these ever needs touching, but it's all a bit tomayto, tomahto.

PR #379, opened 2026-06-04
Add server-side revalidation of total score matching active mods

bdach:
- [x] Depends on https://github.com/ppy/osu-queue-score-statistics/pull/378 for mergeability
- [x] Will want to re-test locally in practice again with old and new clients, it's annoying to do without the above

This PR adds a processor that checks whether for incoming scores `data->'$.total_score_without_mods'` matches `total_score` as defined by the latest score multiplier calculator. This allows an instant cutoff rollout of the new multipliers for all incoming scores without needing to wait on client rollouts.

- If `total_score` does not match `total_score_without_mods`, `total_score` is rewritten based on `total_score_without_mods` and the active mods for the score.
- If `total_score_without_mods` is missing, the score is marked unranked. Production was checked for this, and there have been no incoming legitimate scores without `total_score_without_mods` for more than a year now.

Of particular note, this processor must run *before* everything else, because of data dependencies. Other processors read and do things with `total_score`, such as add it to profile ranked score, or check it for medal purposes, and so on.

The processor is instrumented with a Datadog counter of scores it rewrote. For this counter I had to commit https://github.com/ppy/osu-queue-score-statistics/commit/19c343dd2bd788e232ba5ffc65a65bf6b488e7bf because otherwise it's kind of not possible to log things with the correct Datadog prefix.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/ScoreStatisticsQueueProcessor.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/ScoreStatisticsQueueProcessor.cs
@@ -225,7 +225,7 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor
                                 byte version = item.ProcessHistory.processed_version;
 
                                 foreach (var p in enumerateValidProcessors(score))
-                                    p.RevertFromUserStats(score, userStats, version, conn, transaction, postTransactionActions);
+                                    p.RevertFromUserStats(score, userStats, version, conn, transaction, postTransactionActions, DogStatsd);
                             }
                             else
                             {
@@ -240,7 +240,7 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor
 
                                 try
                                 {
-                                    p.ApplyToUserStats(score, userStats, conn, transaction, postTransactionActions);
+                                    p.ApplyToUserStats(score, userStats, conn, transaction, postTransactionActions, DogStatsd);
                                 }
                                 catch (ProcessingAbortedException abortException)
                                 {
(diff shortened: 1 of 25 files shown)


bdach:
Self-check done, seems to be working as intended as far as my testing goes (tested with old and new lazer client, have not tested score imports, have not tested ddog metrics).

osu.Server.Queues.ScoreStatisticsProcessor/ScoreStatisticsQueueProcessor.cs
```diff
@@ -240,7 +240,7 @@ private void processScore(ScoreItem item, bool force = false)
 
                                 try
                                 {
-                                    p.ApplyToUserStats(score, userStats, conn, transaction, postTransactionActions);
+                                    p.ApplyToUserStats(score, userStats, conn, transaction, postTransactionActions, DogStatsd);
```
peppy, line 243: Kinda ugly ain't it. Considered adding to interface and passing on start, but not much better.

bdach: I almost introduced a context holder object to pass around to these processors instead of spamming further parameters but it would be even more of a disaster on diff size than this already is.

peppy (approved):

peppy:
Intentionally not merging just yet.

osu.Server.Queues.ScoreStatisticsProcessor/ScoreStatisticsQueueProcessor.cs
peppy: Yeah, context would be the other direction. Or just making the `DogStatsd` field static..

bdach: > Or just making the `DogStatsd` field static..

That would require [a PR in another project](https://github.com/ppy/osu-queue-processor/blob/f46702bee601b9d66ff01cf139c5a5470a15080b/osu.Server.QueueProcessor/QueueProcessor.cs#L45) and a NuGet package bump after 🙈 

Legitimately spent like half an hour mulling over the least worst method to do this.

peppy (approved):

PR #380, opened 2026-06-05
Make ES indexing optional when recalculating mod multipliers

peppy:
We'll probably do this as a second pass. Want to get this in to test if it helps with run performance, too.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/RecalculateModMultipliersCommand.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/RecalculateModMultipliersCommand.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Text;
 using System.Threading;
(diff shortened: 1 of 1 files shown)


osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/RecalculateModMultipliersCommand.cs
bdach: Probably remove this now. It appears (as determined via experimentation) that static analysis treats multiple `MemberNotNullWhen` attributes pointing at the same maybe-null member as OR rather than AND.

This is visible when you try to remove the `RunIndexing` guard in https://github.com/ppy/osu-queue-score-statistics/blob/ffa03557c251501464e280b01572aa86eb0f8ca7/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/RecalculateModMultipliersCommand.cs#L195 which should trigger a warning but doesn't, presumably because compiler sees the `DryRun` guard earlier and decides it doesn't need to check further.

There is no way to express this new type of non-null dependency on two fields in conjunction as far as I can tell, so might as well either completely drop these in favour of an assert or a null-ignore, or at least keep just the `RunIndexing` one accepting that it's not entirely correct, but also smidge less dangerously wrong than having both.

bdach (approved):

PR #381, opened 2026-06-10
Test better batching performance by avoiding individual `UPDATE` statements

peppy:


osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/RecalculateModMultipliersCommand.cs
```diff
@@ -174,20 +181,31 @@ public async Task<int> OnExecuteAsync(CancellationToken cancellationToken)
...
+
         private void flush(MySqlConnection conn, bool force = false)
         {
-            int bufferLength = sqlBuffer.Length;
-
-            if (bufferLength == 0)
-                return;
-
-            if (bufferLength > 1024 || force)
+            if (pendingUpdates.Count > FlushSize || force)
```
bdach: This being strict inequality means that e.g. in the default configuration the minimum flush batch size is 1001 and not 1000. Somewhat irrelevant in the large scheme of things given that this isn't supposed to be matching anything.

bdach (approved):
In testing:

- the construction of the `CASE..WHEN..THEN` expression looks correct (no mix-ups between different rows, or mixing up total score with row IDs, when checking against known-correct prior data on a small batch)
- the `IN (ids)` clause looks to encompass only the rows that it is supposed to
- the `CASE..WHEN..THEN` expression exhaustively handles the IDs in the `IN (ids)` clause
- if an ID *was* to be present in the `IN (ids)` clause but missing from the `CASE..WHEN..THEN` expression, the database would stop things:

```
mysql> update scores set total_score = case id when 3 then 47886 end where id = 1;
ERROR 1048 (23000): Column 'total_score' cannot be null
```

Still scary, but probably safe to try.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/RecalculateModMultipliersCommand.cs
peppy: yeah i did notice this and meant to adjust but didn't. will fix before merging just for more round numbers.

PR #383, opened 2026-06-15
Skip mod multiplier updates for rows with suspiciously high result

bdach:
It's a very targeted fix but due to the update batching not much better can be done. If you try-catch the update on the entire batch and log failure, other scores from the batch won't get updated even though they could. So it's either this or very annoying logic to somehow divine which score precisely blew up the update statement and then retroactively excluding it from the batch and then trying the batch again.


peppy (approved):
I'll take it.

PR #385, opened 2026-06-24
Optimise score storage by allowing `total_score_without_mods` to be absent from certain scores

bdach:
- Part of post-https://github.com/ppy/osu/issues/37818 cleanup

## [Exclude scores which are permitted to have total score without mods missing from multiplier validation](https://github.com/ppy/osu-queue-score-statistics/commit/a28f9d6e4d88625b4c7998f7533a63fa5b128aad)

As previously outlined in https://github.com/ppy/osu-queue-score-statistics/pull/374, the only scores that *need* `total_score_without_mods` populated are lazer scores with at least one mod present. To that end, this commit adjusts the validator added in https://github.com/ppy/osu-queue-score-statistics/pull/379 such that it does not mark scores which are allowed to have it unpopulated as unranked, because of not having it populated.

## [Do not populate total score without mods when importing stable scores](https://github.com/ppy/osu-queue-score-statistics/commit/0fdd0f481f4622a4140d515cc1f1d07079fed36f)

Self-explanatory.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor.Tests/ScoreMultiplierValidatorTest.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor.Tests/ScoreMultiplierValidatorTest.cs
@@ -28,11 +28,48 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Tests
         }
 
         [Fact]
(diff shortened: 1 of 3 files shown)


peppy (approved):

bdach:
Warning: for deploy purposes, the queue processor must be deployed ***FIRST*** and the score importers ***SECOND***.

PR #386, opened 2026-07-03
Tidy up pp / user total processing commands

peppy:
See individual commits some extra reasoning. Was also just kind of a mess of commands we never use. Also a precursor to adding better update batching support (same optimisations applied elsewhere using `CASE..WHEN`).

I intentionally left the command structure as-is, even though there's a sub command with only one command under it in two cases now. Just so existing documentation doesn't break. Can adjust if preferred.


osu.Server.Queues.ScoreStatisticsProcessor/Commands/Performance/UserTotals/UpdateAllUserTotalsCommand.cs
```diff
@@ -15,27 +15,37 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Performance.UserTo
     [Command("all", Description = "Updates the total PP of all non-restricted, active users.")]
     public class UpdateAllUserTotalsCommand : PerformanceCommand
     {
+        [Option(Description = "Optional where condition to run against `user_stats` table.", Template = "--where")]
+        public string Where { get; set; } = "1 = 1";
```
bdach, line 19: I guess one weird thing here is that this command processes "all" scores but also has a `Where` filter spec. Not bothered enough about it to demand action.

bdach (approved):

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Performance/UserTotals/UpdateAllUserTotalsCommand.cs
peppy: i'd probably consider renaming the commands if we're ready to break existing documentation (see related mention in OP).

PR #387, opened 2026-07-03
Add support for better batching of score.pp updates

peppy:
- [x] Depends on #386 (to reduce usage of `ProcessScoreAsync` before changing it to not insert itself)

Already production tested.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Performance/Scores/UpdateAllScoresCommand.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Performance/Scores/UpdateAllScoresCommand.cs
@@ -57,18 +57,15 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Performance.Scores
         /// </summary>
         private readonly ConcurrentQueue<MySqlConnection> connections = new ConcurrentQueue<MySqlConnection>();
 
-        private readonly ElasticQueuePusher? elasticQueueProcessor;
+        private ElasticQueuePusher? elasticQueueProcessor;
 
         private readonly ConcurrentBag<ElasticQueuePusher.ElasticScoreItem> elasticItems = new ConcurrentBag<ElasticQueuePusher.ElasticScoreItem>();
 
-        public UpdateAllScoresCommand()
+        protected override async Task<int> ExecuteAsync(CancellationToken cancellationToken)
         {
             if (RunIndexing)
                 elasticQueueProcessor = new ElasticQueuePusher();
-        }
 
-        protected override async Task<int> ExecuteAsync(CancellationToken cancellationToken)
-        {
             using var db = await DatabaseAccess.GetConnectionAsync(cancellationToken);
 
             ulong currentScoreId = From;
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Performance/Scores/UpdateAllScoresCommand.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Performance/Scores/UpdateAllScoresCommand.cs
@@ -173,12 +173,12 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Performance.Scores
                         }
                     }
 
-                    DatabaseHelper.BatchUpdateScoresTable(connection, "scores", "id", "pp", updates);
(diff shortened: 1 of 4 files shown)


osu.Server.Queues.ScoreStatisticsProcessor/Helpers/DatabaseHelper.cs
```diff
@@ -214,5 +217,31 @@ public static bool IsUserRestricted(MySqlConnection db, uint userId, MySqlTransa
...
             return standing.user_type == 1 || standing.user_warnings > 0;
         }
+
+        /// <summary>
+        /// For the given table and column, update all provided (id, value) pairs using CASE..WHEN.
+        /// </summary>
+        /// <remarks>
+        /// This is the most efficient and low overhead method of updating batch values that we've found to date.
+        /// </remarks>
+        public static void BatchUpdateScoresTable<T>(MySqlConnection conn, string tableName, string idColumnName, string valueColumnName, List<(ulong id, T val)> values)
```
bdach, line 227: This method gives me anxiety.

- It is *very* easy to transpose `idColumnName` and `valueColumnName` and in consequence do disastrous things
- Method assumes the arbitrary table it's operating upon has an `ulong` id column

I guess it looks *fine* so up to you if you want to chance it but I am rather concerned. I guess at least I would update all invocations to explicitly name the `idColumnName` and `valueColumnName` parameters so that it doesn't require rider to actually eyeball that the order is correct.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Performance/Scores/UpdateAllScoresCommand.cs
```diff
@@ -57,10 +57,16 @@ public class UpdateAllScoresCommand : PerformanceCommand
...
         private readonly ConcurrentQueue<MySqlConnection> connections = new ConcurrentQueue<MySqlConnection>();
 
-        private readonly ElasticQueuePusher elasticQueueProcessor = new ElasticQueuePusher();
+        private readonly ElasticQueuePusher? elasticQueueProcessor;
 
         private readonly ConcurrentBag<ElasticQueuePusher.ElasticScoreItem> elasticItems = new ConcurrentBag<ElasticQueuePusher.ElasticScoreItem>();
 
+        public UpdateAllScoresCommand()
+        {
+            if (RunIndexing)
```
bdach, line 66: Did you test that this actually works? Is `RunIndexing` going to be populated at ctor time if specified to `true`?

It seems it would be much safer to do this in `ExecuteAsync()`, so I don't get why introduce the ctor at all here.

osu.Server.Queues.ScoreStatisticsProcessor/Helpers/DatabaseHelper.cs
peppy: > Method assumes the arbitrary table it's operating upon has an ulong id column

Actually not too important because it ends up sending as string values anyway. Can be any number type.

osu.Server.Queues.ScoreStatisticsProcessor/Commands/Performance/Scores/UpdateAllScoresCommand.cs
peppy: Yeah this is a fair point, and also what we do elsewhere. I'll move it.

osu.Server.Queues.ScoreStatisticsProcessor/Helpers/DatabaseHelper.cs
```diff
@@ -214,5 +217,31 @@ public static bool IsUserRestricted(MySqlConnection db, uint userId, MySqlTransa
...
+            where T : struct
+        {
+            if (!values.Any())
+                return;
+
+            var statementBuilder = new StringBuilder();
+
+            statementBuilder.AppendLine($"UPDATE {tableName} SET {valueColumnName} = CASE {idColumnName}");
```
bdach, line 235: Identifiers (table, column names) are no longer backtick-quoted here but unless / until it is decided to do something like mass-update `scores.rank` via this method it's probably not an issue.

bdach (approved):

osu.Server.Queues.ScoreStatisticsProcessor/Helpers/DatabaseHelper.cs
peppy: I had to remove it because the `LegacyDatabaseHelper` was also backtick quoting itself. Figure we can add back if/when it's required.

PR #389, opened 2026-07-18
Mark replay deletions in cleanup table

peppy:
This is so resuming a previous run will not need to run for the previous
processed rows again.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/DeleteNonPreservedScoresCommand.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/DeleteNonPreservedScoresCommand.cs
@@ -166,6 +166,8 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Commands.Maintenance
                     if (Verbose)
                         Console.WriteLine($"Deleting replay {score.id}...");
 
+                    bool success;
+
                     if (score.is_legacy_score)
                     {
                         if (score.legacy_score_id < 1)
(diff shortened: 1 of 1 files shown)


osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/DeleteNonPreservedScoresCommand.cs
```diff
@@ -191,10 +193,16 @@ private async Task processCleanupTable(MySqlConnection db, IAmazonS3 s3, Cancell
...
                         var result = await s3.DeleteObjectAsync(S3.REPLAYS_BUCKET, score.id.ToString(CultureInfo.InvariantCulture), cancellationToken);
-                        bool success = await checkS3Success(result);
+                        success = await checkS3Success(result);
                         DogStatsd.Increment("replays_deleted", tags: ["type:new", $"success:{success}"]);
                     }
 
+                    if (success)
+                    {
+                        // Mark scores after replay is deleted so in the case of a resume run, we don't need to call S3.
+                        await db.ExecuteAsync($"UPDATE {scores_cleanup_table} SET has_replay = 0 WHERE score_id = @scoreId", new { scoreId = score.id });
```
bdach: Pretty sure the column name isn't correct here:

```suggestion
                        await db.ExecuteAsync($"UPDATE {scores_cleanup_table} SET has_replay = 0 WHERE id = @scoreId", new { scoreId = score.id });
```

bdach (changes requested):

bdach (approved):
Seems correct.

Have not tested locally as testing locally would involve spending a large amount of time on setup which does not seem like an expedient use of time.

PR #392, opened 2026-07-24
Add support for marking non-preserve scores in realtime

peppy:
This removes the need to repeatedly run `MarkNonPreservedScoresCommand`. If this turns out to run efficiently, this is great news as the batch command is still quite slow (takes days for a full pass).

The result of this change is that users will no longer see bonus PP incorrectly increasing, then getting reset to a lower value when the batch command runs – it will always be in a correct state. It also allows for score cleanup to happen as soon as it can, similar to web-10's handling.

- Test coverage is copied directly from the existing batch process coverage. This command is already known to produce correct results, so hopefully this should be pretty good coverage.
- Could optionally be moved into its own processor if we want. Maybe better for tracking time spent doing the processing? Would need to be run with higher priority than `UserTotalPerformanceProcessor` if so.

Has not yet been performance tested on production, but based on the query execution plan, thing look like they will be okay.

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor.Tests/MarkNonPreservedTests.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor.Tests/MarkNonPreservedTests.cs
@@ -18,11 +18,6 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Tests
         public MarkNonPreservedTests()
         {
             beatmap = AddBeatmap();
-
-            using var db = Processor.GetDatabaseConnection();
-
-            db.Execute("DELETE FROM `multiplayer_playlist_item_scores`");
-            db.Execute("TRUNCATE TABLE `score_pins`");
         }
 
         [Fact]
--- a/osu.Server.Queues.ScoreStatisticsProcessor.Tests/MarkNonPreservedTests.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor.Tests/MarkNonPreservedTests.cs
@@ -1,0 +1,149 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System.Threading.Tasks;
+using Dapper;
+using osu.Server.Queues.ScoreStatisticsProcessor.Models;
+using Xunit;
+
+namespace osu.Server.Queues.ScoreStatisticsProcessor.Tests
+{
+    /// <summary>
+    /// This tests realtime non-preserved marking, handled by `UserTotalPerformanceProcessor`.
+    /// </summary>
+    public class MarkNonPreservedTests : DatabaseTest
+    {
+        private readonly Beatmap beatmap;
+
+        public MarkNonPreservedTests()
+        {
+            beatmap = AddBeatmap();
+
+            using var db = Processor.GetDatabaseConnection();
+
+            db.Execute("DELETE FROM `multiplayer_playlist_item_scores`");
+            db.Execute("TRUNCATE TABLE `score_pins`");
+        }
+
+        [Fact]
+        public void OnlyBestPPAndTotalScoresArePreserved()
+        {
+            SetScoreForBeatmap(beatmap.beatmap_id, s =>
+            {
+                s.Score.id = 1;
+                s.Score.ScoreData.TotalScoreWithoutMods = s.Score.total_score = 800_000;
+                s.Score.pp = 95;
+            });
+            SetScoreForBeatmap(beatmap.beatmap_id, s =>
+            {
+                s.Score.id = 2;
+                s.Score.ScoreData.TotalScoreWithoutMods = s.Score.total_score = 850_000;
+                s.Score.pp = 90;
+            });
+            SetScoreForBeatmap(beatmap.beatmap_id, s =>
+            {
+                s.Score.id = 3;
+                s.Score.ScoreData.TotalScoreWithoutMods = s.Score.total_score = 500_000;
+                s.Score.pp = 85;
+            });
+            WaitForDatabaseState("SELECT COUNT(1) FROM `scores` WHERE `preserve` = 1", 2, CancellationToken);
+            WaitForDatabaseState("SELECT COUNT(1) FROM `scores` WHERE `preserve` = 0", 1, CancellationToken);
+            WaitForDatabaseState("SELECT `preserve` FROM `scores` WHERE `id` = 3", false, CancellationToken);
+        }
+
+        [Fact]
+        public async Task NonBestScoresRemainPreservedIfPinned()
+        {
+            using var db = Processor.GetDatabaseConnection();
+
+            await db.ExecuteAsync("INSERT INTO `score_pins` (`user_id`, `score_id`, `ruleset_id`, `display_order`) VALUES (2, 2, 0, 0)");
+
+            SetScoreForBeatmap(beatmap.beatmap_id, s =>
+            {
+                s.Score.id = 1;
+                s.Score.total_score = 800_000;
+                s.Score.pp = 95;
+            });
+            SetScoreForBeatmap(beatmap.beatmap_id, s =>
+            {
+                s.Score.id = 2;
+                s.Score.total_score = 500_000;
+                s.Score.pp = 85;
+            });
+
+            WaitForDatabaseState("SELECT COUNT(1) FROM `scores` WHERE `preserve` = 1", 2, CancellationToken);
+            WaitForDatabaseState("SELECT COUNT(1) FROM `scores` WHERE `preserve` = 0", 0, CancellationToken);
+            WaitForDatabaseState("SELECT `preserve` FROM `scores` WHERE `id` = 2", true, CancellationToken);
+        }
+
+        [Fact]
+        public async Task NonBestScoresRemainPreservedIfAssociatedWithPlaylistItem()
+        {
+            using var db = Processor.GetDatabaseConnection();
+
+            await db.ExecuteAsync("INSERT INTO `multiplayer_playlist_item_scores` (`user_id`, `playlist_item_id`, `score_id`) VALUES (2, 1, 2)");
+
+            SetScoreForBeatmap(beatmap.beatmap_id, s =>
+            {
+                s.Score.id = 1;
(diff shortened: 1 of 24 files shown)


osu.Server.Queues.ScoreStatisticsProcessor/ProcessorContext.cs
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using MySqlConnector;
+
+namespace osu.Server.Queues.ScoreStatisticsProcessor
+{
+    public record ProcessorContext
+    {
+        public required MySqlConnection Connection { get; init; }
```
bdach, line 10: What's this for? Nothing uses it, it's just set.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserTotalPerformanceProcessor.cs
```diff
@@ -79,6 +83,48 @@ public async Task UpdateUserStatsAsync(UserStats userStats, int rulesetId, MySql
             (userStats.rank_score, userStats.accuracy_new) = UserTotalPerformanceAggregateHelper.CalculateUserTotalPerformanceAggregates(userStats.user_id, scores);
         }
 
+        private async Task markNonPreserved(SoloScore score, MySqlConnection conn, MySqlTransaction transaction, List<Action<ProcessorContext>> postTransactionActions)
+        {
+            List<SoloScore> scores = (await conn.QueryAsync<SoloScore>(
+                """
+                SELECT
+                    s.id, s.beatmap_id, s.ranked,
+                    IF(s.data->'$.mods' IS NULL, '{}', JSON_OBJECT('mods', s.data->'$.mods')) AS data,
```
bdach: Just to confirm, is this an optimization measure to reduce the amount of retrieved data?

osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserTotalPerformanceProcessor.cs
```diff
@@ -79,6 +83,48 @@ public async Task UpdateUserStatsAsync(UserStats userStats, int rulesetId, MySql
...
+                    continue;
+
+                await conn.ExecuteAsync("UPDATE scores SET preserve = 0, unix_updated_at = UNIX_TIMESTAMP() WHERE id = @scoreId", new
+                {
+                    scoreId = s.id
+                }, transaction: transaction);
+
+                postTransactionActions.Add(context =>
+                {
+                    context.ElasticProcessor.PushToQueue(new ElasticQueuePusher.ElasticScoreItem { ScoreId = (long?)s.id });
```
bdach: I am slightly uncomfortable with this and possible implications of it.

The one place I can for sure imagine this backfiring is `HighBeatmapLeaderboardRankEventEmitter` which:
- Executes in `ApplyGlobal()`, so outside of the primary score transaction
- Depends transitively on the state of `preserve` (via `DatabaseHelper.GetUserBestScoreFor()`)
- Depends directly on the state of ES (as it queries web for the score's ranking in the global leaderboard).

So my worry is that doing these things in this way is going to result in some indeterminate failure. Maybe something like repeated first place event scores on setting the same score over and over again, maybe some slightly incorrect leaderboard placements in the emitted events.

I'm not sure how much you care about any of these possibilities. I'm not entirely sure how I'd fix the issue (if there is one here) even.

osu.Server.Queues.ScoreStatisticsProcessor.Tests/MarkNonPreservedTests.cs
```diff
@@ -0,0 +1,149 @@
...
+        private readonly Beatmap beatmap;
+
+        public MarkNonPreservedTests()
+        {
+            beatmap = AddBeatmap();
+
+            using var db = Processor.GetDatabaseConnection();
+
+            db.Execute("DELETE FROM `multiplayer_playlist_item_scores`");
+            db.Execute("TRUNCATE TABLE `score_pins`");
```
bdach: These should be in the base class. Leaving them here is polluting other tests.

I was wondering why `PerformanceProcessorTests.UserDailyRankUpdates` was failing on my machine, and it's because this test is potentially leaving behind stuff in `score_pins` and `multiplayer_playlist_item_scores` that then causes other tests to fail.

osu.Server.Queues.ScoreStatisticsProcessor.Tests/PerformanceProcessorTests.cs
bdach, line 606: Passing comment: It was not exactly intuitive to me why this assertion required no update while the one down below did.

The answer to that is even with the scores getting progressively more pp, the first score will remain preserved regardless on the `total_score` tiebreak combined with `MaxBy(s => s.TotalScore)` apparently returning the first matching value in case of ties.

bdach (commented):
Initial pass. Looks pretty sane in general.

osu.Server.Queues.ScoreStatisticsProcessor/ProcessorContext.cs
peppy: I was actually using it, but ended up not needing it. Can remove it or just let it be, fine either way.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserTotalPerformanceProcessor.cs
peppy: Yeah, correct. Was more important for the batch run, but just copied across from there so should work correctly:

https://github.com/ppy/osu-queue-score-statistics/blob/04e10297df98d36f573f2889bf7ce9c62ad20049/osu.Server.Queues.ScoreStatisticsProcessor/Commands/Maintenance/MarkNonPreservedScoresCommand.cs#L138-L155

osu.Server.Queues.ScoreStatisticsProcessor.Tests/PerformanceProcessorTests.cs
peppy: Yeah, I also updated it then had to figure out why it was not supposed to be.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserTotalPerformanceProcessor.cs
peppy: The other thing to note is that ES can take some time to actually update and give latest results even after an index request, so I don't think that there's any explicit guarantee even before this change.

As half ass as it may sound, I'd probably want to wait for a user noticed failure before even considering how/if to fix this.

bdach (approved):
Generally functionally I do not have anything more to say here.

> Could optionally be moved into its own processor if we want. Maybe better for tracking time spent doing the processing? Would need to be run with higher priority than `UserTotalPerformanceProcessor` if so.

I would probably lean towards separating for multiple reasons:
- Ability to monitor time spent as you say
- Ability to separately turn on/off as required via `DISABLED_PROCESSORS` envvar

Ultimately those are mostly operational upsides so I will leave to you as to how much you care about being able to do those things.

peppy:
Have moved to its own processor, please double-check.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/MarkNonPreservedProcessor.cs
```diff
@@ -0,0 +1,88 @@
...
+
+namespace osu.Server.Queues.ScoreStatisticsProcessor.Processors
+{
+    /// <summary>
+    /// Marks irrelevant scores as non-preserved ahead of calculating totals which may depend on this flag.
+    /// </summary>
+    public class MarkNonPreservedProcessor : IProcessor
+    {
+        // This processor needs to run before user total processor as bonus PP relies on the preserved flag.
+        public const int ORDER = UserTotalPerformanceProcessor.ORDER - 1;
```
bdach: This processor also relies on the incoming score having a pp value assigned. Given

https://github.com/ppy/osu-queue-score-statistics/blob/94d4b997c187aba66e5b9b355a736a4bdfd26ff4/osu.Server.Queues.ScoreStatisticsProcessor/Processors/UserTotalPerformanceProcessor.cs#L23-L24

that is not guaranteed.

These probably need to be daisy-chained: `ScorePerformanceProcessor`, then `MarkNonPreservedProcessor`, then finally `UserTotalPerformanceProcessor`.

(As a side note, this is yet another example wherein the `Order` thing is mighty inconvenient. It would be much easier and simpler to just specify the order as a list in code, though doing that is probably somewhat complicated by private processors existing.)

osu.Server.Queues.ScoreStatisticsProcessor/Processors/MarkNonPreservedProcessor.cs
```diff
@@ -0,0 +1,88 @@
...
+
+        public int Order => ORDER;
+
+        public bool RunOnFailedScores => false;
+
+        public bool RunOnLegacyScores => true;
+
+        // [ruleset_id, [rank_score, count_users_above]]
+        private readonly ConcurrentDictionary<int, MemoryCache> rankScoreIndexPartitionCache =
+            new ConcurrentDictionary<int, MemoryCache>();
```
bdach: ^C^V gone too far

```suggestion
```

bdach (changes requested):

osu.Server.Queues.ScoreStatisticsProcessor/Processors/MarkNonPreservedProcessor.cs
peppy: We could probably just have the private system define its own ordering on top of the base ordering (let it mutate the processor list before it's used).

Something for later.

osu.Server.Queues.ScoreStatisticsProcessor/Processors/MarkNonPreservedProcessor.cs
peppy: [c849d8c](https://github.com/ppy/osu-queue-score-statistics/pull/392/commits/c849d8c7e50871af8faaa6fa1694f032d9edc648)

bdach (approved):
Seems ok

PR #394, opened 2026-07-31
Remove pp column checks from verify scores command

peppy:
New table is source of truth going forward. Old table's column is basically unused at this point. Remaining usages will be migrated away in coming days (probably).

### DIFF
--- a/osu.Server.Queues.ScoreStatisticsProcessor/Helpers/BatchInserter.cs
+++ b/osu.Server.Queues.ScoreStatisticsProcessor/Helpers/BatchInserter.cs
@@ -153,7 +153,7 @@ namespace osu.Server.Queues.ScoreStatisticsProcessor.Helpers
                     bool isRanked = highScore.ShouldPreserve;
 
                     highScore.InsertSql =
-                        $"({highScore.user_id}, {rulesetId}, {highScore.beatmap_id}, {(highScore.replay ? "1" : "0")}, {(highScore.ShouldPreserve ? "1" : "0")}, {(isRanked ? "1" : "0")}, '{referenceScore.Rank.ToString()}', {(highScore.pass ? "1" : "0")}, {referenceScore.Accuracy}, {referenceScore.MaxCombo}, {referenceScore.TotalScore}, '{serialisedScore}', {highScore.pp?.ToString() ?? "null"}, {highScore.score_id}, {referenceScore.LegacyTotalScore}, '{highScore.date:yyyy-MM-dd HH:mm:ss}', {highScore.date.ToUnixTimeSeconds()})";
+                        $"({highScore.user_id}, {rulesetId}, {highScore.beatmap_id}, {(highScore.replay ? "1" : "0")}, {(highScore.ShouldPreserve ? "1" : "0")}, {(isRanked ? "1" : "0")}, '{referenceScore.Rank.ToString()}', {(highScore.pass ? "1" : "0")}, {referenceScore.Accuracy}, {referenceScore.MaxCombo}, {referenceScore.TotalScore}, '{serialisedScore}', {highScore.score_id}, {referenceScore.LegacyTotalScore}, '{highScore.date:yyyy-MM-dd HH:mm:ss}', {highScore.date.ToUnixTimeSeconds()})";
                 }
                 catch (Exception e)
                 {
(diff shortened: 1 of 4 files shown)


bdach (changes requested):
Should also remove

https://github.com/ppy/osu-queue-score-statistics/blob/8f44b2d76feccbc2a9c6b51e357e67de2c39fd39/osu.Server.Queues.ScoreStatisticsProcessor/Models/HighScore.cs#L29

and its only remaining usage in

https://github.com/ppy/osu-queue-score-statistics/blob/653d0963bfe8b5c464bb5340470e4fac1557e4d9/osu.Server.Queues.ScoreStatisticsProcessor/Helpers/BatchInserter.cs#L156

peppy:
> Should also remove
> 
> https://github.com/ppy/osu-queue-score-statistics/blob/8f44b2d76feccbc2a9c6b51e357e67de2c39fd39/osu.Server.Queues.ScoreStatisticsProcessor/Models/HighScore.cs#L29
> 
> and its only remaining usage in
> 
> https://github.com/ppy/osu-queue-score-statistics/blob/653d0963bfe8b5c464bb5340470e4fac1557e4d9/osu.Server.Queues.ScoreStatisticsProcessor/Helpers/BatchInserter.cs#L156

I noticed this one but left it due to mentioned concerns in discussion (this is still used in web-10 for results screen pp-difference display).

If you'd prefer, I can remove the web-10 dependency before we get this change in, then remove completely.

bdach:
> I noticed this one but left it due to mentioned concerns in discussion (this is still used in web-10 for results screen pp-difference display).

I am confused and pretty sure that's not what happens.

1. web-10 inserts a `osu_scores_high_*` row [here](https://github.com/peppy/osu-web-10/blob/60f60a197cc285c7e9d82bd95562ef99f7444883/www/web/osu-submit-20190809.php#L708) without a pp value.
2. The score importer will pick up that row via `BatchInserter` and insert it into `scores`. At that point the `pp` on the row is always `null`.
3. The score inserted in (2) gets pushed for processing to the queue.
4. The pp is written to `scores` *and* `osu_scores_high_*` in https://github.com/ppy/osu-queue-score-statistics/blob/a15d9aeff4b4ac913120164499d76e402fc05e10/osu.Server.Queues.ScoreStatisticsProcessor/Processors/ScorePerformanceProcessor.cs#L55-L72
5. If steps (2)-(4) execute quick enough, web-10 picks up the pp value inserted in (4) [here](https://github.com/peppy/osu-web-10/blob/60f60a197cc285c7e9d82bd95562ef99f7444883/www/web/osu-submit-20190809.php#L774).

Unless something on production isn't configured as I'd expect it to be at this point, the presence of `pp` in `BatchInserter` seems to be providing nothing of value. In other words, I would expect that every score picked up from the old tables by `BatchInserter` has null pp on it, so there should be no reason not to remove it.

Although if you'd rather not touch it for now I guess that's fine, as long as it eventually goes (along with the aforementioned pp update in old tables).

peppy:
Yeah you're right sorry. Misread the code.

osu.Server.Queues.ScoreStatisticsProcessor/Helpers/BatchInserter.cs
```diff
@@ -174,7 +174,7 @@ private async Task run(HighScore[] scores)
 
             bool first = true;
             StringBuilder insertBuilder = new StringBuilder(
-                "INSERT INTO scores (`user_id`, `ruleset_id`, `beatmap_id`, `has_replay`, `preserve`, `ranked`, `rank`, `passed`, `accuracy`, `max_combo`, `total_score`, `data`, `pp`, `legacy_score_id`, `legacy_total_score`, `ended_at`, `unix_updated_at`) VALUES ");
+                "INSERT INTO scores (`user_id`, `ruleset_id`, `beatmap_id`, `has_replay`, `preserve`, `ranked`, `rank`, `passed`, `accuracy`, `max_combo`, `total_score`, `data`, klegacy_score_id`, `legacy_total_score`, `ended_at`, `unix_updated_at`) VALUES ");
```
bdach:
```suggestion
                "INSERT INTO scores (`user_id`, `ruleset_id`, `beatmap_id`, `has_replay`, `preserve`, `ranked`, `rank`, `passed`, `accuracy`, `max_combo`, `total_score`, `data`, `legacy_score_id`, `legacy_total_score`, `ended_at`, `unix_updated_at`) VALUES ");
```

bdach (approved):

